Issue #42, opened 2026-08-20
[Chore]: Apply Prettier repo-wide to settle the quote/semicolon drift from #37


SrLampi1001:
## Motivation

PR #39 (Tailwind config) ran `prettier --write` on the three files it touched. That converted `vite.config.ts` from the project's earlier style (single quotes, no trailing semicolons) to Prettier's defaults (double quotes, semicolons). The rest of the repo — `README.md`, `CONTRIBUTING.md`, `docs/**`, `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `src/main.tsx`, `tsconfig.json` — has the same drift, inherited from PR #37.

Today this is silent: `pnpm format:check` fails, but every PR so far has either ignored it or fixed only its own files. The result is that the next person to touch `vite.config.ts` will inherit a format choice that was forced by Prettier, not chosen by a person.

If we let it ride, every new contributor pays a "wait, when did we switch to double quotes?" tax on every drive-by edit, and `format:check` keeps being a non-gate that nobody trusts.

## Proposed Approach

One-shot cleanup PR, scoped **only** to formatting changes:

1. Run `pnpm format` (or `prettier --write .`) once on the whole tree.
2. Commit as `chore(format): apply prettier to the whole tree (#N)`.
3. Branch from `main` after #39 lands so the diff stays scoped to file contents, not new conflicts.
4. Optionally run `prettier --check .` in CI after this lands — that depends on #10 (quality gate) wiring the script up.

If we want the Prettier config itself to match what the existing files already do (single quotes, no trailing semis), that's a separate decision worth filing as its own follow-up — it should be made deliberately, not as a side effect of a cleanup PR.

## Definition of Done

- [ ] `pnpm format:check` exits 0 on `main`.
- [ ] No content-meaningful diffs in the PR (every change is whitespace, quotes, semicolons, or line length).
- [ ] PR title follows Conventional Commits (`chore(format): …`).
- [ ] PR references this issue with `Closes #N`.
- [ ] `pnpm build` still succeeds (sanity check; formatting shouldn't affect output, but cheap to verify).
- [ ] `tsc --noEmit -p tsconfig.app.json` still exits 0.

## Risk

Low. Mechanical, idempotent, easy to revert (`git revert` of a single squash commit returns the tree to its pre-PR state).

## Affected Area

- [ ] tooling (scripts, CI, linters)

## Additional Notes

[...]

PR #39, opened 2026-08-20
chore(ui): configure tailwind css + theme tokens (#4)

SrLampi1001:
## Description

Configures Tailwind CSS 4 with the project's design-contract tokens (CSS-first `@theme`) and ships the `cn()` class-merge utility. This is the **Paso 6** layer from `docs/guides/baseline.md` and is the foundation that every UI component will build on.

Closes #4

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional change, just code improvement)
- [x] Chore / Maintenance (dependency upgrade, tooling, docs)
- [ ] Performance improvement
- [ ] Test only

## What Changed

- **`vite.config.ts`** — register the `@tailwindcss/vite` plugin between `react()` and `tsconfigPaths()`.
- **`src/index.css`** — replace the placeholder comment with `@import "tailwindcss"`, the `@theme` block defining semantic colors (surface, surface-raised, ink, ink-muted, brand, danger), catalog status colors (released, unreleased, unknown), `text-rating` + line-height, `spacing-touch`, `radius-card`, `aspect-poster`, and a `prefers-reduced-motion` fallback that neutralizes animations and transitions.
- **`src/presentation/lib/cn.ts`** — new two-line utility that composes `clsx` + `tailwind-merge` so component code can merge classes without string concatenation (the silent "two paddings in the same chain" footgun the baseline calls out).

## Affected Layers

- [ ] `domain/` — pure TypeScript, no framework imports
- [ ] `application/` — use cases and ports
- [ ] `infrastructure/` — HTTP, storage, API
- [x] `presentation/` — React, routes, hooks

### Before

N/A — no visual change yet. `src/App.tsx` still renders `<h1>Cineteca</h1>`; Tailwind utilities and `cn()` are wired but not yet consumed by any component (that is the next step in the baseline chain).

### After

N/A — see above.

## How to Test

1. Pull the branch: `git fetch && git checkout chore/tailwind-config`
2. `pnpm install` (no new dependencies; sanity check)
3. `pnpm build` — `dist/assets/index-*.css` should be ~6–7 kB (Tailwind v4 reset + the `@theme` definitions). Confirmed locally: 6.64 kB / 2.10 kB gzipped.
4. `npx tsc --noEmit -p tsconfig.app.json` — should exit 0. Confirmed locally.
[...]

Issue #38, opened 2026-08-20
[Chore]: Establish a version-pinning convention for package.json


SrLampi1001:
## Description

`package.json` currently has two coexisting styles for version ranges, because the baseline deps (#1, #2 — `react: ^19.2.8`, `typescript: ~6.0.2`, `vite: ^8.2.0`) use semver ranges while the install-deps PR (#3, #36 — `axios: 1.19.0`, `eslint: 10.8.1`, …) pinned every new entry exactly per the spec.

This was flagged as item #3 in the [review on #36](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/pull/36#pullrequestreview-4983934665):

> Version-pinning inconsistency in the same `package.json`. … the resulting mixed-style file is worth a follow-up governance issue (e.g. "loosen the spec to allow `^` for runtime deps, keep `~` for tooling"). Worth opening as a separate chore so the conversation doesn't live forever on this PR.

This issue is that follow-up.

## Tasks

- [ ] Decide on a single version-pinning convention for this repo (see Options below).
- [ ] Document the convention in `CONTRIBUTING.md` once PR #37 lands. Until then, this issue is the canonical reference.
- [ ] Either loosen the exact pins from #3 to the chosen range, or leave them exact and standardize on exact for future deps.
- [ ] Update the baseline spec (project guide / Cinética BaseLine.md) to reflect the convention so the next person writing a deps PR isn't guessing.

## Options to consider

| Convention | Pros | Cons |
|---|---|---|
| **`^` for runtime, `~` for tooling, exact for nothing** | Auto-patch/minor updates; one rule per dep type | Mixing styles inside one file; `typescript-eslint` peer-dep constraints (#3's `eslint-plugin-jsx-a11y@6.10.2` only lists `^3..^9`) mean we sometimes need `~` for a *tool* to keep ESLint 10 green |
| **All exact** (current #3 state for new deps) | Reproducible; matches the spec verbatim; no surprise updates | No auto-patches; every transitive bugfix becomes a manual PR; lockfile churns forever |
| **All `^`** (closest to `package.json` defaults) | Auto-patches and minor bumps for everything | Risky for tooling (ESLint major updates break configs) and for libraries with breaking minors (Zod 4 had `z.prettifyError` renames) |
[...]

Issue #10, opened 2026-08-20
[Chore]: Quality gate (verify.sh + husky + CI)


3105jero:
## Description
This is the most important task of the baseline. The gate runs the same checks locally and in CI — a `main` with a warning is not allowed.

## Tasks
- [ ] Create `scripts/verify.sh` with `--quick` and `--full` modes
- [ ] Quick mode: format check, lint, types
- [ ] Full mode: quick + tests + build + dependency check
- [ ] Create `scripts/check-versions.sh` to detect deprecated dependencies
- [ ] Configure husky pre-commit with `pnpm lint-staged` + `bash scripts/verify.sh --quick`
- [ ] Configure husky pre-push with `bash scripts/verify.sh --full`
- [ ] Add `lint-staged` config in `package.json` for `*.ts,*.tsx` and `*.json,*.css,*.md`
- [ ] Create `.github/workflows/ci.yml` running `verify.sh --full` on push and PR
- [ ] Configure branch protection on `main` requiring the CI job

## Time Budget
- Quick gate: under 10 seconds
- Full gate: under 90 seconds

> A slow gate gets `--no-verify`'d on a Friday at 10pm. That's how you lose.

## Acceptance Criteria
- [ ] Trying to commit an `any` typed variable is rejected
- [ ] `bash scripts/verify.sh --full` is green
- [ ] CI workflow exists and runs the same script
- [ ] `main` is protected

## Reference
[Cinética BaseLine.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cin%C3%A9tica%20BaseLine.md) — Step 12.

## Dependencies
- Depends on #9 App skeleton

PR #43, opened 2026-08-20
chore(gate): configure eslint, prettier, and architecture rules (#5)

### LINKED ISSUE #5: [Chore]: Configure ESLint + architecture rules
## Description
This is the file that turns "Clean Architecture" from a diagram into an executable rule. Domain must remain pure TypeScript.

## Tasks
- [ ] Create `eslint.config.js` (NOT `.ts` — saves a loader)
- [ ] Wire `js`, `globals`, `typescript-eslint`, `react-hooks`, `jsx-a11y`, `@tanstack/eslint-plugin-query`, `eslint-config-prettier`
- [ ] Apply `strictTypeChecked` + `stylisticTypeChecked`
- [ ] Add rules: `no-explicit-any: error`, `no-unnecessary-condition: error`, `consistent-type-imports: error`, `switch-exhaustiveness-check: error`, `no-console: error`
- [ ] Add the **dependency rule** for `src/domain/**`: no React, no axios, no `@tanstack/*`, no outer layer imports
- [ ] Add the **dependency rule** for `src/application/**`: no React, no axios, no outer layer imports
- [ ] Add the **axios rule** for everywhere except `src/infrastructure/http/**`
- [ ] Create `.prettierrc.json` with `singleQuote: true`, `semi: true`, `printWidth: 100`, `trailingComma: all`
- [ ] Run the 60-second demo: import React in domain, see linter fail, remove it

## Acceptance Criteria
- [ ] `pnpm lint` passes with zero warnings
- [ ] `pnpm format:check` passes
- [ ] Importing React in `src/domain/` is rejected with the rule message
- [ ] Importing axios outside `infrastructure/http/` is rejected

## Reference
[Cinética BaseLine.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cin%C3%A9tica%20BaseLine.md) — Step 7.

## Dependencies
- Depends on #4 Tailwind theme

SrLampi1001:
## Description

Adds the **quality-gate enforcement layer** to the repo: ESLint flat config (with the architecture dependency rule that turns "Clean Architecture" from a diagram into an executable rule) and a Prettier config that matches the project's existing style. After this lands, `pnpm lint` runs green on the current tree and rejects the violations described in `docs/guides/baseline.md` Paso 7.

> **Review feedback addressed in commit `1d9749e`** (on top of the original `19d95a4`):
> - Translated all five linter messages and the in-file section comments from Spanish to English to match the rest of the repo (`CONTRIBUTING.md`, `docs/`, READMEs, PR descriptions).
> - Replaced the inaccurate "override" claim in the catch-all-axios comment with the correct "merge" description. ESLint flat config **merges** configs at the rule level — it does not replace them.
> - Scoped the `@typescript-eslint/no-deprecated` disable to the specific line that uses `tseslint.config()`, so any future deprecation in this file still gets caught.
> - Added a `See PR #6` lineage comment on the `*.spec.{ts,tsx}` block so its connection to the Vitest config is explicit.
>
> The squash merge will collapse both commits into one on `main`; the historical record on the PR is for reviewability.

Closes #5

## Type of Change

- [ ] Performance improvement
- [ ] Test only

## What Changed

- **`eslint.config.js`** — new flat config. Composes `@eslint/js` recommended, `typescript-eslint` `strictTypeChecked` + `stylisticTypeChecked`, `eslint-plugin-react-hooks` (`flat/recommended`), `eslint-plugin-jsx-a11y` (`flatConfigs/recommended`), `@tanstack/eslint-plugin-query` (`flat/recommended`), and `eslint-config-prettier` (always last). Strict-type-checked rules are scoped to `src/**/*.{ts,tsx}` only, so `eslint.config.js` itself is never type-checked (avoids unsafe-`*` noise on `import.meta.dirname`).
- **`eslint.config.js`** — adds the **dependency rule**:
  - `src/domain/**` rejects `react`, `react-*`, `axios`, `@tanstack/*`, `react-hook-form`, and any `@/presentation/*`, `@/infrastructure/*`, `@/application/*` import — with the messages *"Domain does not depend on frameworks."* and *"Dependencies point inward."*
  - `src/application/**` rejects `@/presentation/*`, `@/infrastructure/*`, `axios`, `react`, `react-*` — with the messages *"Application defines interfaces; infrastructure implements them, not the other way around."* and *"Application does not know how the data travels."*
  - Catch-all `src/**/*.{ts,tsx}` excluding `src/infrastructure/http/**` rejects `axios` with *"Only src/infrastructure/http may import axios."*
- **`.prettierrc.json`** — new. `singleQuote: true, semi: true, printWidth: 100, trailingComma: "all"`. Matches the baseline guide and the style every existing file was written in.
- **`package.json`** — adds `@eslint/js@10.0.1` to devDependencies. It is a transitive of `eslint`, but the baseline imports it directly so it must be a direct dep too. Pinned (no `^`) per the project's existing devDependency pinning convention. The pinning convention itself is tracked in **#38** and is intentionally not resolved here.
- **`pnpm-lock.yaml`** — refreshed by `pnpm add -D`.
- **`src/main.tsx`** — replaces `document.getElementById('root')!` with a proper null check (`if (!rootElement) throw new Error('Root element #root not found.')`). The `!` was a pre-existing violation that `strictTypeChecked`'s `no-non-null-assertion` (error level) catches.
- **`src/index.css`** — `@import "tailwindcss"` → `@import 'tailwindcss'` to match `singleQuote: true`.
- **`src/presentation/lib/cn.ts`** — re-quoted to single quotes (was Prettier default double-quote from #39).
- **`vite.config.ts`** — re-quoted to single quotes for the same reason.

## Affected Layers

### Before

`pnpm lint` errors with:
```
ESLint couldn't find an eslint.config.(js|mjs|cjs) file.
```
…and `pnpm format:check` is meaningless (no `.prettierrc.json`).

### After

`pnpm lint` → exit 0, no warnings, no output. `pnpm format:check` → green for all files this PR touches; the wider repo still has drift inherited from #37, addressed by #42.

## How to Test

1. Pull the branch: `git fetch && git checkout chore/eslint-config`
2. `pnpm install` (pulls `@eslint/js`)
3. `pnpm lint` → exits 0
4. `pnpm format:check eslint.config.js .prettierrc.json src/main.tsx src/index.css vite.config.ts src/presentation/lib/cn.ts` → "All matched files use Prettier code style!"
5. **Run the 60-second dependency-rule demo** from `docs/guides/baseline.md` Paso 7:
   ```bash
   mkdir -p src/domain/shared
   echo 'import { useState } from "react"; export const x = useState;' > src/domain/shared/bad.ts
   pnpm lint
   # Expected: error "Domain does not depend on frameworks."
   rm -rf src/domain/shared
   ```
   All four variants in the baseline guide have been verified locally:
   - `react` import in domain → "Domain does not depend on frameworks." ✅
   - `axios` import in domain → "Domain does not depend on frameworks." ✅
   - `@/presentation/*` import in domain → "Dependencies point inward." ✅
   - `axios` import in `src/infrastructure/http/` → no error (allowed) ✅
   - Bonus: `axios` import in `src/presentation/components/foo.ts` → "Only src/infrastructure/http may import axios." ✅

## Tests

- [ ] I added unit tests for the new logic
- [ ] I updated existing tests that no longer apply
- [x] I verified the manual test plan above (all 4 dependency-rule cases pass)
- [ ] Coverage has not decreased (no tests in this PR; the Vitest config lands in #6)

## Quality Gate

- [x] `pnpm format:check` passes (for the files this PR touches)
- [x] `pnpm lint` passes (zero warnings) — **this is the gate this PR enables**
- [x] `pnpm check-types` passes (run via `npx tsc --noEmit -p tsconfig.app.json`; the npm script is the broken pre-existing one flagged in #35)
- [ ] `pnpm test` passes (coverage thresholds met) — **not run**: no tests in this PR; Vitest config lands in #6.
- [x] `pnpm build` succeeds — bundle sizes unchanged from #39 (`dist/assets/index-*.css` is still 6.64 kB / 2.10 kB gzipped).
- [ ] `./scripts/check-versions.sh` reports no deprecated deps — **not run**: script lands in #10.

## Checklist

- [x] I have added comments only where the logic is non-obvious (block-level comments in `eslint.config.js` explain the flat-config merge behavior and the deprecation disable)
- [x] I have updated the documentation (README, copy, etc.) if needed — `docs/guides/baseline.md` already documents this step

## Deployment Notes

_None._ No env vars, no migrations, no config toggles. CI will pick up the new lint step automatically when #10 wires the gate.

## Reviewer Focus

- **`eslint.config.js` (the dependency-rule blocks)** — flat config **merges** configs at the rule level; it does not replace them. The catch-all axios block (lines ~50-70) declares `paths: [axios]`; the domain and application blocks below add their `patterns: [...]` to the same merged rule for the files they match. Empirically `patterns` is evaluated first when both could match — that's why `import 'axios'` in a domain file gets the domain-flavored message rather than the catch-all's. The block comment in the file spells out the three observed outcomes.
- **`eslint.config.js` (`eslint-disable-next-line @typescript-eslint/no-deprecated`)** — `tseslint.config()` is deprecated in favor of `defineConfig()` from eslint core, but eslint 10.8.1 does not yet export `defineConfig`. The disable is now scoped to the single line that uses it.
- **`src/main.tsx`** — the `!` → null-check swap is a real behavior change at runtime: previously, if `#root` was missing the app would crash with a `TypeError: Cannot read properties of null (reading 'appendChild')` deep inside React's internals; now it crashes with a clear `Error: Root element #root not found.` from our code. Both fail, but the second one tells the developer what to fix.
- **`@eslint/js@10.0.1` pin** — matches the eslint@10.8.1 major. Bump together when eslint bumps. The exact-pin vs `^`-range decision is tracked in **#38**.

## Pre-existing issues this PR surfaces (NOT blockers)

- **`pnpm format:check` still fails on `CONTRIBUTING.md`, `README.md`, `docs/**/*.md`, `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `tsconfig.json`** — pre-existing drift inherited from #37. Filed as **#42**.
- **`tsconfck@3.1.6` deprecation warning** at install time — pre-existing, flagged in snapshot §6.5.
- **`eslint-plugin-jsx-a11y@6.10.2` peer-dep warning** (`eslint ^3..^9` vs installed `10.8.1`) — pre-existing, snapshot §6.5; runtime/lint unaffected.


3105jero (approved):
This is the PR that turns "Clean Architecture" from a diagram into an executable rule, and it's well-executed. The architecture dependency rule is correct, the messages are at least directionally useful, and the 60-second demo in the PR body is exactly the kind of thing that lets a future contributor verify the gate is doing what we think it's doing. Approving with a few non-blocking comments.

### What works

- **The dependency rule is structurally correct.** Catch-all axios ban (`src/**` minus `infrastructure/http/**`) + domain rule + application rule, with the right messages and the right targets.
- **Practical UX is right.** In a domain file importing `react`, you get the domain message ("El dominio no depende de frameworks"); importing `axios` you get the catch-all axios message. The PR's claim that the more specific rules "override" the catch-all is loose (see below), but the effective behavior — right message for the right context — is what we want.
- **`src/main.tsx` null-check swap** is a real win. The `!` non-null assertion would have been flagged by `strictTypeChecked`'s `no-non-null-assertion: error` rule, and the explicit error message is a genuine improvement over React's internal `Cannot read properties of null (reading 'appendChild')` stack. The PR body explains this clearly.
- **Prettier config matches the existing style** (`singleQuote: true`, `semi: true`). The re-quoting in `src/index.css`, `src/presentation/lib/cn.ts`, `vite.config.ts` is no longer a drive-by — it's the Prettier config landing, which is what should happen.
- **`projectService: true`** with `tsconfigRootDir: import.meta.dirname` is the modern typescript-eslint way to avoid per-tsconfig config in `parserOptions.project`. ✓
- **The 60-second demo** in the PR body is exactly the right thing to ship — anyone can verify the gate is doing what we think.
- **`prettier` last** in the config array. ✓
- **Honest Quality Gate** with pre-existing prettier drift explicitly filed as #42.

1. **"Override" claim in the comment + PR body is technically inaccurate.** The code comment says: *"Las reglas más específicas debajo sobreescriben esta"* and the PR body repeats it. ESLint flat config **merges** configs at the rule level — it doesn't override. For `src/domain/**`, the effective `no-restricted-imports` rule has **both** `paths: [axios]` (from catch-all) and `patterns: [react, axios, ...]` (from domain). They coexist; neither replaces the other. The practical behavior is fine because ESLint reports the first match it finds, but the wording should be "merge" or "layer on top of", not "override". Worth a 30-second comment fix in both the ESLint config and the PR description.

2. **Spanish error messages in an otherwise-English project.** CONTRIBUTING.md, docs/, README, PR descriptions — all English. The linter messages here are in Spanish ("El dominio no depende de frameworks", "Las dependencias apuntan hacia dentro", "La aplicación no sabe cómo viajan los datos"). Two options, both defensible:
   - Translate to English to match the rest of the project.
   - Keep Spanish deliberately and update CONTRIBUTING to acknowledge the bilingual convention.
   What's currently shipped reads like it slipped in, not like a decision. Pick one. (If the team is bilingual and Spanish messages are easier to internalize during code review, that's a legitimate reason to keep them — but document it.)

3. **`eslint-disable @typescript-eslint/no-deprecated` is file-wide.** The comment correctly identifies that `tseslint.config()` is the deprecated entry point, but the disable applies to the whole file. A line-specific disable (`// eslint-disable-next-line @typescript-eslint/no-deprecated`) above the `tseslint.config(...)` call would scope it tighter and let the rule catch any other deprecation that gets added later. Minor but the right shape.

### Nits (not blocking)

4. **`@eslint/js@10.0.1` pinned exact (no `^`)** while every other devDep in `package.json` uses semver ranges. This is the same inconsistency flagged in the PR #36 review — every new "pinned exact" dep widens the gap. Worth a one-off governance issue: "pick a single pinning convention and apply it retroactively". Filing that as a chore would close the loop on every future PR having to make this decision.

5. **`{ files: ['**/*.spec.{ts,tsx}'], rules: { ... } }` block.** This is defensive — there are no `*.spec.ts` files yet. Fine to leave in, but flagging because it's untested code. The moment a real spec file lands in #6 / Vitest work, this becomes load-bearing; the moment someone renames the convention, it silently stops working. A `// See PR #6 / #43` comment would make the lineage explicit.

6. **The `paths` vs `patterns` distinction in `no-restricted-imports` does mean axios reports the catch-all message in a domain file** (since `paths` checks fire before `patterns`). Cosmetic, but if you wanted the domain-flavored message for axios-in-domain too, you could move axios into the domain block's `patterns` group and drop the catch-all — at the cost of repeating "axios" in two places.

### Verdict

Approve. Item 1 is the one I'd actually fix in this PR (one-line comment edit), because it'll be the first thing a contributor reads when they hit the rule for the first time and wonder why "specific rules" supposedly "override". The rest are follow-ups or convention decisions that don't need to gate this merge.

This unblocks the whole feature chain (#19–#22 domain types, #23 HTTP client, etc.). Approving now is more valuable than holding it for the nits.

Issue #22, opened 2026-08-20
[Feature]: NoData type for absences


3105jero:
## Description
TMDB returns `0` or empty strings for unknown data. Translate these to an explicit "no data" at the boundary — the type must tell the truth.

## Tasks
- [ ] Create `NoData` type as part of every nullable domain field
- [ ] Translation rules:
  - `0` or `""` for budget → `NoData`
  - `0` for vote count → `NoData`
  - `null` or `""` for synopsis → `NoData`
- [ ] UI components must handle the `NoData` case explicitly
- [ ] All translations happen at the validation edge, never deep in the app

## The Rule
> No `0` that means "I don't know".

## Acceptance Criteria
- [ ] A missing budget shows "no data", never `$0`
- [ ] A missing poster shows a placeholder, never a broken icon
- [ ] A missing synopsis in Spanish offers the English version with a notice

## Parent
#11 — Domain modeling (states, money, rating)

Issue #11, opened 2026-08-20
[Feature]: Domain modeling (states, money, rating)


3105jero:
## Description
Model the domain in pure TypeScript with **no framework dependencies**. This is the foundation — no UI is built until the domain is modeled and tested.

### Movie State (Discriminated Union)
- [ ] Model movie lifecycle as a tagged union: `Released`, `Unreleased`, `Unknown`
- [ ] Each branch carries exactly the data it has — no nullable fields shared across branches
- [ ] `switch` over the union must be exhaustive (`switch-exhaustiveness-check` enforces it)

### Rating Reliability
- [ ] Model rating reliability as a tagged union: `Consolidated` (>= N votes), `Early` (low votes), `Unknown`
- [ ] Each branch carries the rating + the count that justifies it

### Money Type
- [ ] Create `Money` type: integer amount in the smallest currency unit + ISO currency code
- [ ] **Never** allow a plain number to represent money in the type system
- [ ] Build formatters that delegate to `Intl.NumberFormat` (currency code from the data, locale from the browser)

### Absences
- [ ] TMDB returns `0` or empty strings for unknown data — translate these to **explicit "no data"** at the boundary
- [ ] No `null`/`undefined` ambiguity inside the domain

## The Two Rules That Go In the README
1. **No number representing money outside the `Money` type.**
2. **No `0` that means "I don't know".**

## Acceptance Criteria
- [ ] Domain covers all TMDB states with exhaustive switches
- [ ] **100% coverage** in `domain/` (it's pure and cheap to test)
- [ ] No imports from React, axios, or `@tanstack/*`
- [ ] Linter rejects any domain file importing React

## Reference
[Cineteca.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cineteca.md) — "El dominio: estados, dinero y ausencias".

## Dependencies
- Depends on #10 Quality gate

Issue #9, opened 2026-08-20
[Chore]: Build app skeleton (main, providers, routes, layout)


3105jero:
## Description
Build the app skeleton: entry point, providers, router, and layout. **No functionality yet** — just the scaffold that runs with the gate green.

## Tasks
- [ ] Create `src/main.tsx` — imports CSS, mounts React in `<StrictMode>`, wraps in providers
- [ ] Create `src/presentation/providers/app-providers.tsx` — QueryClientProvider + ErrorBoundary (TanStack Query devtools in dev only)
- [ ] Create `src/presentation/routes/router.tsx` — root route, root layout, 404 route
- [ ] Create `src/presentation/routes/root-layout.tsx` — header, footer with TMDB attribution, `<Outlet />`
- [ ] Use TanStack Query devtools only in development

## Three Decisions Locked In Now
1. **React StrictMode** from day one — better to discover double-mount today than day 5
2. **ErrorBoundary** from first commit — a render error never leaves the screen blank
3. **TMDB attribution in the footer** from day one — required by TMDB terms, easily forgotten

## Acceptance Criteria
- [ ] `pnpm dev` serves the skeleton with header + footer
- [ ] An unknown route shows the 404 page
- [ ] Console is clean
- [ ] TMDB attribution is visible in the footer

## Reference
[Cinética BaseLine.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cin%C3%A9tica%20BaseLine.md) — Step 11.

## Dependencies
- Depends on #8 Env validation

PR #45, opened 2026-08-20
chore(test): configure vitest, msw, and jsdom polyfills (#6)

### LINKED ISSUE #6: [Chore]: Configure Vitest + MSW + jsdom polyfills
## Description
Set up the test runner, MSW for network mocking, and the jsdom polyfills the app needs.

## Tasks
- [ ] Create `vitest.config.ts` with `environment: 'jsdom'`, `globals: true`, coverage config
- [ ] Create `vitest.setup.ts` with `@testing-library/jest-dom/vitest`, MSW lifecycle, `matchMedia` and `ResizeObserver` stubs
- [ ] Create `src/test/msw/server.ts` with `setupServer()` (no handlers yet)
- [ ] Add a smoke test that mounts the app shell and checks the header renders
- [ ] Keep coverage thresholds **commented** until there's domain code to cover

## Acceptance Criteria
- [ ] `pnpm test` runs and the smoke test passes
- [ ] MSW logs `request not handled` for un-simulated URLs (this is a feature, not a bug)
- [ ] `pnpm test:watch` works

## Reference
[Cinética BaseLine.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cin%C3%A9tica%20BaseLine.md) — Step 8.

## Dependencies
- Depends on #5 ESLint rules

SrLampi1001:
## Description

Sets up the **test runner, MSW for network mocking, and the jsdom polyfills the app needs**. After this lands, `pnpm test` runs a smoke test green, MSW is wired with `onUnhandledRequest: 'error'` so future tests that hit un-simulated URLs will fail loudly instead of silently going to the network, and the spec-file dependency-rule escape hatch from PR #43 has its first real file to exercise.

> **Review feedback addressed in commit `ce9a328`** (on top of the original `d71ded0`):
> - Added the `@tailwindcss/vite` plugin to `vitest.config.ts`. It was missing — the vitest config mirrors the dev plugins, and `vite.config.ts` has had `tailwindcss()` since #4. The smoke test passed regardless because `<h1>Cineteca</h1>` uses no Tailwind classes, but the first feature PR that uses `bg-surface` / `text-ink` / `min-h-touch` / `rounded-card` would have rendered unstyled and the failure would have looked like a component bug rather than a missing config. Caught and fixed before any feature PR lands.
>
> Five non-blocking review items filed as separate issues:
> - **#46** — extract `tsconfig.spec.json` to scope jest-dom matchers to spec files
> - **#47** — replace the `ResizeObserver` double-cast with a typed stub or document the why
> - **#48** — drop the redundant `cleanup()` call from `vitest.setup.ts` (auto-handled in `@testing-library/react@16`)
> - **#49** — pick and document a Vitest convention (globals vs explicit imports)
> - **#50** — update or delete `src/App.spec.tsx` when the app shell lands in #9
>
> The squash merge will collapse both commits into one on `main`; the historical record on the PR is for reviewability.

Closes #6

## Type of Change

- [ ] Performance improvement
- [ ] Test only

## What Changed

- **`vitest.config.ts`** (root) — Vitest config with `environment: 'jsdom'`, `globals: true`, `setupFiles: ['./vitest.setup.ts']`, v8 coverage with reporter `text` and `html`, includes `src/**/*.{ts,tsx}`, excludes spec files / `src/main.tsx` / `.d.ts` / `src/test/**`. Coverage **thresholds are intentionally commented out** until the domain layer lands (per the baseline guide — turning them on with an empty scaffold only teaches the gate how to negotiate). **Plugins** mirror `vite.config.ts`: `react(), tailwindcss(), tsconfigPaths()`.
- **`vitest.setup.ts`** (root) — `@testing-library/jest-dom/vitest` for matchers, MSW `server.listen({ onUnhandledRequest: 'error' })` before all tests, `server.resetHandlers()` + `cleanup()` after each, `server.close()` after all. Plus the two jsdom polyfills the theme and virtualizer will need: a fake `window.matchMedia` and a stub `ResizeObserver`.
- **`src/test/msw/server.ts`** — `setupServer()` with no handlers. Handlers are added per-feature when they're needed, alongside the feature that depends on them.
- **`src/App.spec.tsx`** — smoke test. Renders `<App />` and asserts the `Cineteca` heading is in the document. First real consumer of the `*.spec.{ts,tsx}` block from PR #43's eslint config.
- **`tsconfig.app.json`** — extends `types` with `@testing-library/jest-dom/vitest`. Without this, the matcher type augmentation (`toBeInTheDocument` etc.) is invisible to spec files because `vitest.setup.ts` lives at the repo root and isn't in `tsconfig.app.json`'s `include`. The augmentation needs to be loaded by TypeScript at the project level.

## Affected Layers

- [x] `presentation/` — React, routes, hooks

### Before

`pnpm test` errors with:
```
vitest: command not found (vitest is in devDependencies but not configured)
```

### After

```
$ pnpm test

 RUN  v4.1.11 /home/cohorte5/Documentos/ssr_documents/Team-Centinela/Project-save-me-of-the-riwi
      Coverage enabled with v8

 Test Files  1 passed (1)
      Tests  1 passed (1)
   Duration  1.49s
```

## How to Test

1. Pull the branch: `git fetch && git checkout chore/vitest-config`
2. `pnpm install` (no new deps — all are already on `main` from #3)
3. `pnpm test` → "Test Files 1 passed (1), Tests 1 passed (1)"
4. `pnpm test:watch` → watcher starts, smoke test runs and re-runs on save
5. (Optional, MSW smoke) Add a temporary spec file that does `fetch('/api/whatever')` — `onUnhandledRequest: 'error'` will fail the test instead of letting it through to the network.

## Tests

- [ ] I added unit tests for the new logic
- [x] I updated existing tests that no longer apply (no prior tests)
- [x] Coverage has not decreased (no coverage threshold set; v8 report shows only `cn.ts` at 0% because the smoke test doesn't import it)

## Quality Gate

- [x] `pnpm format:check` passes (for the files this PR touches)
- [x] `pnpm lint` passes (zero warnings) — first spec file under the new gate, validates the `*.spec.{ts,tsx}` block works
- [x] `pnpm check-types` passes (run via `npx tsc --noEmit -p tsconfig.app.json`)
- [x] `pnpm test` passes — **this is the gate this PR enables**
- [x] `pnpm build` succeeds — bundle sizes unchanged from #5 (`dist/assets/index-*.css` is still 6.66 kB / 2.10 kB gzipped).
- [ ] `./scripts/check-versions.sh` reports no deprecated deps — **not run**: script lands in #10.

## Checklist

- [x] I have added comments only where the logic is non-obvious (the `onUnhandledRequest: 'error'` rationale, the coverage-threshold comment, the jsdom-polyfill rationale)
- [x] I have updated the documentation (README, copy, etc.) if needed — `docs/guides/baseline.md` already documents this step

## Deployment Notes

_None._ No env vars, no migrations, no config toggles. CI will pick up the new test step automatically when #10 wires the gate.

## Reviewer Focus

- **`vitest.config.ts` plugins** — now `react(), tailwindcss(), tsconfigPaths()`, mirroring `vite.config.ts`. The fix landed in this PR per the review's "1-line fix I'd push into this PR" suggestion.
- **`tsconfig.app.json` `types`** — adding `@testing-library/jest-dom/vitest` makes the matcher augmentation globally available in `src/`. This is intentional for now; **#46** tracks the extraction of `tsconfig.spec.json` to scope it properly.
- **`vitest.setup.ts` `onUnhandledRequest: 'error'`** — strict by default. The first time a feature lands with a fetch call that doesn't have an MSW handler, the test will fail loudly. This is the **feature**, not a bug — per the baseline guide.
- **Coverage thresholds** — commented out per the baseline guide's explicit instruction. When the domain layer lands (#11–#22) and there's real code to cover, this is the place to uncomment.
- **`src/App.spec.tsx`** — minimal. Just renders the existing `<App />` (which is still `<h1>Cineteca</h1>`). When the proper app shell lands in #9, **#50** tracks updating or deleting this test.

## Pre-existing issues this PR surfaces (NOT blockers)

- Same as #5: `pnpm format:check` still fails on `CONTRIBUTING.md`, `README.md`, `docs/**/*.md`, `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `tsconfig.json` — drift inherited from #37. Filed as **#42**.
- `tsconfck@3.1.6` deprecation warning at install time — pre-existing, snapshot §6.5.
- `eslint-plugin-jsx-a11y@6.10.2` peer-dep warning — pre-existing, snapshot §6.5; runtime/lint unaffected.

### DIFF
--- a/src/App.spec.tsx
+++ b/src/App.spec.tsx
@@ -1,0 +1,10 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import App from './App';
+
+describe('App', () => {
+  it('renders the project name as a heading', () => {
+    render(<App />);
+    expect(screen.getByRole('heading', { name: /Cineteca/i })).toBeInTheDocument();
+  });
+});
--- a/src/test/msw/server.ts
+++ b/src/test/msw/server.ts
@@ -1,0 +1,5 @@
+import { setupServer } from 'msw/node';
+
+// Starts empty on purpose: every handler is added alongside the feature
+// that needs it, with real API responses as the baseline.
+export const server = setupServer();
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -9,7 +9,7 @@
     "target": "es2023",
     "lib": ["ES2023", "DOM"],
     "module": "esnext",
-    "types": ["vite/client"],
+    "types": ["vite/client", "@testing-library/jest-dom/vitest"],
     "allowArbitraryExtensions": true,
     "skipLibCheck": true,
 
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,0 +1,25 @@
+import { defineConfig } from 'vitest/config';
+import react from '@vitejs/plugin-react';
+import tsconfigPaths from 'vite-tsconfig-paths';
+
+export default defineConfig({
+  plugins: [react(), tsconfigPaths()],
+  test: {
+    environment: 'jsdom',
+    globals: true,
+    setupFiles: ['./vitest.setup.ts'],
+    coverage: {
+      provider: 'v8',
+      reporter: ['text', 'html'],
+      include: ['src/**/*.{ts,tsx}'],
+      exclude: ['src/**/*.spec.{ts,tsx}', 'src/main.tsx', 'src/**/*.d.ts', 'src/test/**'],
+      // Thresholds are intentionally kept off until the domain layer lands.
+      // Turning them on with an empty scaffold only teaches the gate how to
+      // negotiate with you.
+      // thresholds: {
+      //   lines: 80, functions: 80, branches: 80, statements: 80,
+      //   'src/domain/**/*.ts': { lines: 100, functions: 100, branches: 100, statements: 100 },
+      // },
+    },
+  },
+});
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,9 +1,10 @@
 import { defineConfig } from 'vitest/config';
 import react from '@vitejs/plugin-react';
+import tailwindcss from '@tailwindcss/vite';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
 export default defineConfig({
-  plugins: [react(), tsconfigPaths()],
+  plugins: [react(), tailwindcss(), tsconfigPaths()],
   test: {
     environment: 'jsdom',
     globals: true,
--- a/vitest.setup.ts
+++ b/vitest.setup.ts
@@ -1,0 +1,36 @@
+import '@testing-library/jest-dom/vitest';
+import { cleanup } from '@testing-library/react';
+import { afterAll, afterEach, beforeAll, vi } from 'vitest';
+import { server } from './src/test/msw/server';
+
+// `onUnhandledRequest: 'error'` is what makes MSW useful: a request that
+// nobody simulated blows up the test instead of going to the real network.
+beforeAll(() => {
+  server.listen({ onUnhandledRequest: 'error' });
+});
+afterEach(() => {
+  server.resetHandlers();
+  cleanup();
+});
+afterAll(() => {
+  server.close();
+});
+
+// jsdom does not implement either of these, but the theme and the
+// virtualizer expect them.
+Object.defineProperty(window, 'matchMedia', {
+  writable: true,
+  value: vi.fn().mockImplementation((query: string) => ({
+    matches: false,
+    media: query,
+    onchange: null,
+    addEventListener: vi.fn(),
+    removeEventListener: vi.fn(),
+    dispatchEvent: vi.fn(),
+  })),
+});
+globalThis.ResizeObserver = class {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+} as unknown as typeof ResizeObserver;


3105jero (approved):
Strong test-infra PR. MSW lifecycle is correct, the polyfills are the right ones with the right rationale, the smoke test passes, and `onUnhandledRequest: 'error'` is exactly the strict-by-default stance the baseline calls for. Approving with one real concern and a handful of non-blocking comments.

### What works

- **MSW lifecycle** is textbook: `listen` before all, `resetHandlers` + `cleanup` after each, `close` after all. Order is correct — `resetHandlers` before `cleanup` so a handler that fires during cleanup still gets cleared.
- **`onUnhandledRequest: 'error'`** is the strict-by-default stance the baseline explicitly calls for. PR body explains the rationale in two lines.
- **Polyfills are the right two**: `matchMedia` (theme detection will need it) and `ResizeObserver` (`@tanstack/react-virtual` will need it). The PR body explains *why* each one exists, not just *that* it exists.
- **Coverage `exclude`** correctly excludes spec files, main entry, declaration files, and the test infrastructure itself. The thresholds are commented out with the right justification.
- **`tsconfig.app.json` `types`** addition is the right pragmatic choice — global type augmentation is simpler than introducing a `tsconfig.spec.json`, and it's type-only so it's a no-op at runtime.
- **Smoke test is minimal on purpose.** Asserts the heading by accessible role and accessible name, which is exactly the Testing Library philosophy.
- **Quality Gate** is honest about what was verified locally vs not (`scripts/check-versions.sh` lands in #10).

### Real concern (worth fixing in this PR)

1. **`vitest.config.ts` is missing the `@tailwindcss/vite` plugin.** `vite.config.ts` has `plugins: [react(), tailwindcss(), tsconfigPaths()]`. The vitest config mirrors the dev plugins — that's the whole point — but it stops at `plugins: [react(), tsconfigPaths()]`. The smoke test passes because `<App />` is still just `<h1>Cineteca</h1>` with no Tailwind classes. The first feature PR that uses `bg-surface` or `text-ink` will render unstyled and the test will fail in a way that looks like a component bug, not a config bug.
   The fix is one line:
   ```ts
   import tailwindcss from '@tailwindcss/vite';
   // ...
   plugins: [react(), tailwindcss(), tsconfigPaths()],
   ```
   I'd land it now while the vitest config is being touched, rather than waiting for the first feature PR to discover the gap. Cheap fix, prevents a confusing failure later.

### Should look at (small, follow-up OK)

2. **`tsconfig.app.json` `types: [..., '@testing-library/jest-dom/vitest']`** makes the matcher augmentation globally available in `src/`. It works because the augmentation is type-only, but it means **any** file in `src/` could write `expect(...)` and TypeScript wouldn't complain — the matchers are visible everywhere. The cleaner pattern is a `tsconfig.spec.json` that extends `tsconfig.app.json` and scopes the types to spec files only. PR body acknowledges this; fine as a follow-up, but worth filing the chore.

3. **`globalThis.ResizeObserver` cast** uses `as unknown as typeof ResizeObserver` — a double cast that bypasses type checking. The cast is necessary because jsdom's `ResizeObserver` interface has more members than the stub provides, but the smell is real. Cleaner:
   ```ts
   class ResizeObserverStub implements ResizeObserver {
     observe() { /* noop */ }
     unobserve() { /* noop */ }
     disconnect() { /* noop */ }
   }
   globalThis.ResizeObserver = ResizeObserverStub;
   ```
   Or just accept the double cast and add a one-line comment explaining "intentional polyfill, real type is wider than the stub". Pick one and document.

4. **`cleanup()` from `@testing-library/react`** — with `@testing-library/react@16.x`, `cleanup` is auto-registered when `afterEach` is available globally (which Vitest always provides). The explicit `import { cleanup }` + `afterEach(() => { ... cleanup() })` is redundant. Calling it twice is a no-op so this is harmless, but it's dead code that future readers will wonder about.

### Nits (not blocking)

5. **`globals: true`** in `vitest.config.ts` enables global `describe`/`it`/`expect`. The smoke test then does `import { describe, expect, it } from 'vitest'` — redundant but harmless. Pick one convention (either rely on globals or always import) and apply it consistently across spec files. The current state means the first contributor who writes a spec without imports gets a working test, and the second who writes one with imports also gets a working test, and nobody knows which is the house style.

6. **`src/App.spec.tsx` is going to need updating when #9 lands** (real app shell → router). PR body flags this. Worth filing a tracking issue now so the test doesn't silently rot.

### Verdict

Approve. Item 1 is a 1-line fix that I'd push into this PR (it's the same PR that introduces the config it's missing). Items 2-6 are follow-ups that don't gate this merge.

This unblocks the actual feature work — domain types (#19-22), HTTP client (#23), components (#13-17) — all of which can now be tested from day one.

Issue #44, opened 2026-08-20
[Chore]: Adopt ADR documentation system + AGENTS.md for decision and language enforcement


SrLampi1001:
## Motivation

The repo is now large enough that **decisions are starting to get lost**. Three concrete examples from the last 24 hours of baseline work:

1. **The lint dependency-rule messages** (PR #43) were committed in Spanish because they were copy-pasted from `docs/guides/baseline.md` Paso 7. The rest of the repo is in English (`CONTRIBUTING.md`, `docs/`, READMEs, every PR body, every commit message). The mismatch was only caught in PR review. There was no document saying "the repo's spoken language is English" — it was implicit, and a contributor (me, in this case) reasonably interpreted the Spanish baseline guide as "this is what we want".
2. **The `@eslint/js@10.0.1` exact pin** (PR #43) was added without a `^` range because the existing devDeps from #3 were pinned exact. The choice between exact-pin and semver-range is a policy decision that affects every future deps PR. It's tracked in **#38**, but #38 is one specific decision; what we don't have is a *place* where future decisions of this kind can land.
3. **The flat-config merge vs. override behavior** (PR #43) is a non-obvious behavior of ESLint v10 that future contributors will hit the first time they try to extend the dependency rule. The why-is-this-config-shaped-this-way answer is currently buried in a comment in `eslint.config.js`. That works for the next reader of the file, but it doesn't help the next contributor who's *deciding* whether to deviate from the pattern.

These are not failures — they're the predictable cost of building in the open without a decision log. The fix is the same one every codebase eventually arrives at: **a lightweight ADR (Architecture Decision Record) system** + a **machine-readable file** that surfaces them to the next contributor and to the next AI agent.

> **Silent requirement from the assignment.** The Cineteca assignment requires ADR-style documentation of architectural decisions; the requirement was communicated orally and was not written into the assignment text. This issue picks it up so it doesn't get forgotten again.

### 1. Adopt a lightweight ADR system under `docs/adr/`

Use the format popularized by [Michael Nygard's "Documenting Architecture Decisions"](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions-michael-nygard) — one Markdown file per decision, named `NNNN-short-title.md`, numbered monotonically, immutable once accepted (new decisions supersede via a new ADR that references the old one).

Each ADR has four sections:

```markdown
# N. Title

- **Status:** Proposed | Accepted | Superseded by #N+1
[...]

PR #55, opened 2026-08-20
chore(config): validate env at startup with zod (#8)

SrLampi1001:
## Description

Closes #8

Validates the three TMDB environment variables at startup with Zod. If any
required variable is missing or malformed, the app dies at module load with
a clear, field-tagged error message — not three screens later with an
obscure auth error.

```
src/config/env.ts          Zod schema + parse-on-load + export env
src/config/env.spec.ts     5 specs covering happy path + 3 failure modes
src/main.tsx               Imports env at the entry point so the validation fires
.env.example               Committed template (placeholder values, no secrets)
src/config/.gitkeep        Removed — directory is now populated.
```

## Type of Change

- [x] Chore / Maintenance (tooling, startup wiring)

## What Changed

- **`src/config/env.ts` (new)** — Zod schema validating:
  - `VITE_TMDB_READ_TOKEN` — required, `min(40)` (TMDB v3 tokens are 64-char hex; v4 JWTs are longer).
  - `VITE_TMDB_API_BASE` — required URL, default `https://api.themoviedb.org`.
  - `VITE_TMDB_IMAGE_BASE` — required URL, default `https://image.tmdb.org/t/p`.

  Exports `env` (the parsed result) and `Env` (the inferred type). On parse failure, throws `Invalid environment configuration:\n<z.prettifyError output>` — the field name and reason are visible.

- **`src/config/env.spec.ts` (new)** — 5 specs:
  1. Happy path with all three values set — `env` matches input.
  2. Defaults applied when API base URLs are omitted.
  3. Overrides honored when API base URLs are provided.
  4. Missing `VITE_TMDB_READ_TOKEN` throws `/Invalid environment configuration/`.
  5. Short token and non-URL base throw field-tagged errors.

  Uses `vi.resetModules` + `vi.stubEnv` + dynamic `await import('./env')` to exercise the real parse-on-load behavior, not just the schema in isolation.

- **`src/main.tsx` (modified)** — imports `env` at the entry point. Without this, Vite tree-shakes the unused `env.ts` module and the validation never fires. The import is referenced by a dev-mode `console.warn` so the import is not dropped:
  ```ts
  if (import.meta.env.DEV) {
    console.warn(`[cineteca] env validated against ${env.VITE_TMDB_API_BASE}`);
  }
  ```
  Production cost: the schema parse (idempotent, ~zero work) and the unused-import warning is dead-code-eliminated.

[...]

PR #51, opened 2026-08-20
chore(gate): apply prettier repo-wide and ignore the lockfile (#42)

SrLampi1001:
## Description

Closes #42

Pre-#37 files (`CONTRIBUTING.md`, `README.md`, `docs/**`, `package.json`, the
issue templates, `.github/dependabot.yml`, `tsconfig.json`) were never
formatted against the `.prettierrc.json` contract adopted in #43. Without
this sweep, `pnpm format:check` reports drift on files that have not been
touched since the bootstrap, so the gate is not honest for incoming PRs.

This is the exact one-shot cleanup the #39 reviewer suggested
([comment 1](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/pull/39#pullrequestreview-4984374163)).

## Type of Change

- [x] Chore / Maintenance (dependency upgrade, tooling, docs)

## What Changed

- `.prettierignore` (new) — exclude `pnpm-lock.yaml` so future `pnpm install`
  produces no cosmetic churn unrelated to actual dependency changes.
- `CONTRIBUTING.md` — quote/semicolon/line-length reformat.
- `README.md` — same.
- `docs/architecture.md`, `docs/quality-gate.md`, `docs/tech-stack.md`,
  `docs/guides/baseline.md`, `docs/guides/project-guide.md` — same.
- `package.json` — was committed minified to one line in #1; expanded to
  multi-line per the current `printWidth: 100` config. No value changes.
- `tsconfig.json` — same.
- `.github/dependabot.yml`, `.github/ISSUE_TEMPLATE/{bug_report,chore,feature_request}.yml`
  — same.

No content-meaningful diffs. Every change is whitespace, quotes,
semicolons, or line-length wrap.

## Affected Layers

- [x] None — tooling and docs only.

## Screenshots / Recordings

N/A — no UI change.

### Before

N/A

### After

N/A

## How to Test

1. Pull the branch: `git fetch && git checkout chore/prettier-repo-wide`
2. Run `pnpm install --frozen-lockfile`
3. Run `pnpm format:check` — must report "All matched files use Prettier code style!"
4. Run the full gate (see Quality Gate below).

## Tests

- [x] I verified the manual test plan above
- [x] Coverage has not decreased (still 50% — only `cn.ts` is covered, same as `main`)

## Quality Gate

- [x] `pnpm format:check` passes — `All matched files use Prettier code style!`
- [x] `pnpm lint` passes — 0 errors, 0 warnings
[...]

Issue #27, opened 2026-08-20
[Chore]: Enable branch protection on `main` and `develop` (linear history OFF)


SrLampi1001:
## Goal

Enable GitHub branch protection rules on `main` and `develop` so that nothing reaches those branches without a reviewed pull request and a green quality gate. This is the foundation of the governance model and must land before the other governance issues (#28, #29, #30, #32) become meaningful.

> **Important correction from a review.** A previous draft required linear history on `main` and `develop`. That setting literally forbids merge commits and would force the `develop ↔ main` round-trip to use squash — which breaks the long-lived relationship (see #29). The rule below has linear history **off** on both branches.

### `main`

- **Require a pull request before merging**
  - Required approvals: **1**
  - Dismiss stale pull request approvals when new commits are pushed
  - Require review from Code Owners (once a `CODEOWNERS` file is added)
- **Require status checks to pass before merging**
  - Required check: `gate` (from `.github/workflows/ci.yml`)
  - Require branches to be up to date before merging
- **Require conversation resolution before merging**
- **Do not require linear history** — merge commits from `release/*` and `hotfix/*` are required by #29
- **Do not allow force pushes**
- **Do not allow deletions**
- **Include administrators** (rules apply to repo admins too)
- **Allow auto-merge**

### `develop`

- **Require a pull request before merging**
  - Required approvals: **1**
  - Dismiss stale pull request approvals when new commits are pushed
- **Require status checks to pass before merging**
  - Required check: `gate`
  - Require branches to be up to date before merging
- **Require conversation resolution before merging**
- **Do not require linear history** — same reason as `main`
- **Do not allow force pushes**
- **Do not allow deletions**
- **Do not include administrators** (admins can fast-forward `develop` while iterating on the next release)

## Required merge strategies on this repo

Both "Allow squash merging" and "Allow merge commits" are enabled (per #29). The rules above do not change that; they only enforce the PR + review + CI surface. The choice of which button to click per PR lives in `CONTRIBUTING.md` under "Merge strategy — hybrid, on purpose."

## Why

The repo currently has **no branch protection on `main`** and no `develop` branch at all. Anyone with push access can land code without a review or a passing CI check. That is incompatible with the assignment's gate rule ("nada entra a `main` sin el gate") and with the audit trail a graded deliverable requires.

## Definition of Done

- [ ] `main` is protected with the rules above.
- [ ] `develop` branch exists and is protected with the rules above.
[...]

PR #60, opened 2026-08-20
chore(gate): add quality gate with verify.sh, husky, lint-staged, and CI

### LINKED ISSUE #10: [Chore]: Quality gate (verify.sh + husky + CI)

SrLampi1001:
# Pull Request

## Description

Wires the project's quality gate end-to-end so a `main` with a warning or a deprecated dependency is no longer possible — the most important task of the baseline. Local hooks and CI run the **same** script (`scripts/verify.sh`); that single fact is the point of the whole quality system.

Closes #10

## Type of Change

- [ ] Performance improvement
- [ ] Test only

## What Changed

- **`scripts/verify.sh`** (new, executable) — the single gate script. Two modes:
  - `--quick` runs `pnpm format:check` → `pnpm lint` (zero warnings) → `pnpm check-types`. Target: < 10 s. Used by the pre-commit hook.
  - `--full`  runs `--quick` plus `pnpm test` (Vitest with coverage) → `pnpm build` → `bash scripts/check-versions.sh --gate`. Target: < 90 s. Used by the pre-push hook and CI.
  - The first failing step stops the script, the script names which step, and the script exits non-zero. There is no `|| true` anywhere.
- **`scripts/check-versions.sh`** (new, executable) — prints a table of installed vs latest for every direct dependency, flags any deprecated package, and with `--gate` exits non-zero if anything is deprecated (which is what verify.sh uses). Without `--gate` it's a human-readable report.
- **`.husky/pre-commit`** (new, executable) — runs `pnpm lint-staged` (ESLint `--fix` + Prettier on staged files only) followed by `bash scripts/verify.sh --quick` (the entire project). Two different scopes; both are needed.
- **`.husky/pre-push`** (new, executable) — runs `bash scripts/verify.sh --full`. The whole gate fires before anything leaves the machine.
- **`package.json`**:
  - Adds `"prepare": "husky"` so `pnpm install` wires `core.hooksPath = .husky` for every developer who clones fresh.
  - Adds a top-level `lint-staged` config for `*.{ts,tsx}` (eslint --fix + prettier --write) and `*.{json,css,md}` (prettier --write).
- **`.github/workflows/ci.yml`** (new) — runs the same gate on every push to `main` and every pull request. Ubuntu + Node 22 + pnpm via corepack + `pnpm install --frozen-lockfile` + `bash scripts/verify.sh --full`, with `VITE_TMDB_READ_TOKEN` injected from a repository secret. The `gate` check is the only thing protecting `main` once branch protection is wired (Deployment Notes below).

## Affected Layers

- [x] None — tooling and infrastructure only (scripts, hooks, CI, package.json).

## Screenshots / Recordings

N/A — no UI change.

### Before

```bash
$ bash scripts/verify.sh --quick
bash: scripts/verify.sh: No such file or directory
```

…and CI runs nothing on pull requests; a `main` with a warning can land at any time.

### After

```bash
$ bash scripts/verify.sh --quick

▶ [1/5] Format — Prettier
$ prettier --check .
Checking formatting...
All matched files use Prettier code style!

▶ [2/5] Linter — ESLint, zero warnings
$ eslint . --max-warnings 0

▶ [3/5] Types — tsc --noEmit
$ tsc --noEmit

✔ QUICK GATE GREEN in 27s
```

## How to Test

1. Pull the branch: `git fetch && git checkout chore/gate-add-quality-gate`
2. Run `pnpm install` — this also wires Husky (`prepare` script).
3. Run `bash scripts/verify.sh --quick` → exits 0.
4. Run `bash scripts/verify.sh --full` → exits 0 (format, lint, types, tests, build, no deprecated deps).
5. **Run the 60-second "any" demo** (the acceptance criterion from #10):
   ```bash
   echo 'export const x: any = 1;' > src/domain/_demo_any.ts
   pnpm lint
   # Expected: error "@typescript-eslint/no-explicit-any — Unexpected any."
   rm src/domain/_demo_any.ts
   ```
   This is the rule wired in #43 and made unforgeable by the `--max-warnings 0` flag in the lint script that verify.sh runs.
6. Open the PR — CI runs the same `verify.sh --full` and posts a green `Quality gate` check.

## Tests

- [ ] I added unit tests for the new logic — **N/A**: the gate is a bash wrapper around `pnpm` scripts that already have unit tests. `scripts/verify.sh` and `scripts/check-versions.sh` are shell scripts and not unit-tested in this project (see #44 for the ADR question of whether shell tests would change that).
- [x] I verified the manual test plan above (`--quick` and `--full` both green locally; "any" demo rejects the commit).
- [x] Coverage has not decreased — still 94.73 % lines, 75 % branches on the 19 statements the project has.

## Quality Gate

- [x] `pnpm format:check` passes — `All matched files use Prettier code style!`
- [x] `pnpm lint` passes — 0 errors, 0 warnings (this is the gate this PR enables)
- [x] `pnpm check-types` passes — `tsc --noEmit` exits 0
- [x] `pnpm test` passes — 6 files, 20 tests, 94.73 % lines, 75 % branches, 87.5 % funcs, 94.73 % stmts
- [x] `pnpm build` succeeds — `built in 865ms`, `dist/assets/index-BZqrTi9p.js 370.50 kB / 113.75 kB gzipped`
- [x] `./scripts/check-versions.sh --gate` reports no deprecated deps — all 41 deps `ok` (no `DEPRECATED` rows)

## Checklist

- [x] My code follows the project's architecture rules (no domain code touched)
- [x] I have used the codebase's existing patterns and utilities (mirrors the baseline guide Paso 12 step for step, in English to match the rest of the repo per the #43 review feedback)
- [x] I have added comments only where the logic is non-obvious (block-level comments at the top of each script explain the contract and who uses it)
- [x] I have updated the documentation (README, copy, etc.) if needed — `docs/quality-gate.md` already documents this step; no edits needed there

## Deployment Notes

After this lands on `main`, the repo admin (`@3105Jero` per branch-protection ownership) needs to enable the branch-protection rule on `main` that requires the `Quality gate` check before merge:

**Settings → Branches → Branch protection rules → `main`:**

| Setting                                        | Value                                                  |
| ---------------------------------------------- | ------------------------------------------------------ |
| Require a pull request before merging          | **enabled**                                            |
| Required approving reviews                      | 1                                                      |
| Require status checks to pass before merging   | **enabled**, require "Quality gate"                    |
| Require linear history                         | **disabled** (see `CONTRIBUTING.md` — linear history forbids merge commits and would force the squash-only problem back) |
| Include administrators                         | enabled                                                |

This is the only piece of #10 that this PR cannot do from code; it's a GitHub UI step. Once it's set, the rule from `CONTRIBUTING.md` that "every commit on `main` is releasable" is mechanical.

## Reviewer Focus

@3105Jero — three things to look at:

1. **`scripts/verify.sh` step ordering and the `--quick` exit path.** Steps 1–3 always run. Steps 4–5 only run in `--full`. The `--quick` exit prints its own green message and returns 0 without running `check-versions.sh`. Both branches land in a clean exit; the script names every step it tried. Worth eyeballing that the failure semantics are obvious to anyone running it.
2. **`scripts/check-versions.sh` `--gate` mode.** It exits non-zero if any package row reads `DEPRECATED`. It does **not** exit non-zero for rows reading `has $latest` (an outdated-but-not-deprecated upgrade). That's intentional: only `deprecated` is a gate-stopper; outdated-but-supported versions are a Dependabot concern, not a gate concern. If you want outdated to also block, change line 60 (`[ "$installed" != "$latest" ]`) to also `exit 1`. PRs against this behavior are welcome, but worth confirming the intent first.
3. **`package.json` `lint-staged`** runs `eslint --fix --max-warnings 0` then `prettier --write` on staged files only. The `verify.sh --quick` step that runs immediately after checks the **entire** project — they are different scopes, both needed, and the order (lint-staged → verify) is important: lint-staged fixes the staged files first so the project-wide check has the fixes already in place.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,35 @@
+name: ci
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  gate:
+    name: Quality gate
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: pnpm
+
+      - name: Enable Corepack
+        run: corepack enable
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run quality gate
+        run: bash scripts/verify.sh --full
+        env:
+          VITE_TMDB_READ_TOKEN: ${{ secrets.VITE_TMDB_READ_TOKEN }}
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,2 @@
+pnpm lint-staged
+bash scripts/verify.sh --quick
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,1 @@
+bash scripts/verify.sh --full
--- a/package.json
+++ b/package.json
@@ -16,7 +16,17 @@
     "format:check": "prettier --check .",
     "check-types": "tsc --noEmit",
     "test": "vitest run --coverage",
-    "test:watch": "vitest"
+    "test:watch": "vitest",
+    "prepare": "husky"
+  },
+  "lint-staged": {
+    "*.{ts,tsx}": [
+      "eslint --fix --max-warnings 0",
+      "prettier --write"
+    ],
+    "*.{json,css,md}": [
+      "prettier --write"
+    ]
   },
   "dependencies": {
     "@hookform/resolvers": "5.9.1",
--- a/scripts/check-versions.sh
+++ b/scripts/check-versions.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# =============================================================================
+# check-versions.sh — Compare installed packages against the latest stable
+# release on the registry and flag deprecated ones.
+# With --gate: exits non-zero if any package is deprecated (used by verify.sh).
+# Without --gate: prints a table for human consumption.
+# =============================================================================
+set -euo pipefail
+cd "$(dirname "$0")/.."
+
+GATE="${1:-}"
+FAILED=0
+
+deps=$(node -p "const p=require('./package.json');Object.keys({...p.dependencies,...p.devDependencies}).join('\n')")
+
+printf '%-40s %-14s %-14s %s\n' "PACKAGE" "INSTALLED" "LATEST" "STATUS"
+
+while IFS= read -r dep; do
+  [ -z "$dep" ] && continue
+  installed=$(node -p "try{require('$dep/package.json').version}catch(e){'?'}" 2>/dev/null || echo '?')
+  latest=$(pnpm view "$dep" version 2>/dev/null || echo '?')
+  deprecated=$(pnpm view "$dep" deprecated 2>/dev/null || true)
+  status="ok"
+  if [ -n "$deprecated" ]; then
+    status="DEPRECATED"
+    FAILED=1
+  elif [ "$installed" != "$latest" ] && [ "$installed" != "?" ] && [ "$latest" != "?" ]; then
+    status="has $latest"
+  fi
+  printf '%-40s %-14s %-14s %s\n' "$dep" "$installed" "$latest" "$status"
+done <<< "$deps"
+
+if [ "$GATE" = "--gate" ] && [ "$FAILED" -ne 0 ]; then
+  echo "✖ Deprecated dependencies detected. Replace them with their documented successor." >&2
+  exit 1
+fi
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# =============================================================================
+# verify.sh — THE quality gate.
+# Contract: exit 0 ⇒ format + lint with zero warnings + strict types +
+#           tests green + build succeeds + no deprecated dependencies.
+# Used by: .husky/pre-commit (--quick), .husky/pre-push (--full), and CI (--full).
+# Never skipped, never weakened "to make it pass": fix the code instead.
+# =============================================================================
+set -euo pipefail
+cd "$(dirname "$0")/.."
+
+MODE="${1:---full}"
+case "$MODE" in
+  --quick | --full) ;;
+  *) echo "usage: verify.sh [--quick|--full]" >&2; exit 2 ;;
+esac
+
+BLUE='\033[1;34m'
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+NC='\033[0m'
+
+step() { printf '\n%b▶ %s%b\n' "$BLUE" "$1" "$NC"; }
+fail() { printf '\n%b✖ GATE RED — %s%b\n' "$RED" "$1" "$NC" >&2; exit 1; }
+
+START=$(date +%s)
+
+step "[1/5] Format — Prettier"
+pnpm format:check || fail "files are not formatted (run: pnpm format)"
+
+step "[2/5] Linter — ESLint, zero warnings"
+pnpm lint || fail "the linter found problems"
+
+step "[3/5] Types — tsc --noEmit"
+pnpm check-types || fail "type errors"
+
+if [ "$MODE" = "--quick" ]; then
+  printf '\n%b✔ QUICK GATE GREEN in %ss%b\n' "$GREEN" "$(( $(date +%s) - START ))" "$NC"
+  exit 0
+fi
+
+step "[4/5] Tests — Vitest with coverage"
+pnpm test || fail "tests failed or coverage below threshold"
+
+step "[5/5] Production build"
+pnpm build || fail "the build failed"
+
+bash scripts/check-versions.sh --gate || fail "a dependency is deprecated"
+
+printf '\n%b✔ FULL GATE GREEN in %ss%b\n' "$GREEN" "$(( $(date +%s) - START ))" "$NC"


3105jero (approved):
This is the foundational PR — the gate that every subsequent PR runs through. It's well-structured, the script is clean, and the contract ("local = CI = same script") is the right architectural call. Approving with one real concern and a handful of nits.

### What works

- **Single source of truth for the gate.** `scripts/verify.sh` is the only script that knows the step ordering. Husky hooks, lint-staged, and CI all call into it. That single fact — local and CI cannot diverge — is the whole point.
- **`set -euo pipefail` + `|| fail` pattern.** Errors propagate correctly; the explicit `fail` calls print a clear "GATE RED" line with the failing step's name. Anyone running the script sees exactly which step failed.
- **Step ordering matches the spec**: format → lint → types → tests → build → dependency check. The fastest/cheapest failures run first.
- **`MODE="${1:---full}"` default.** With no args, full mode runs. The `case` block rejects anything else with exit code 2 — `verify.sh typo` is a fast, loud failure, not a silent one.
- **`pnpm lint-staged` runs before `verify.sh --quick` in pre-commit.** Lint-staged fixes the staged files in isolation; verify then checks the whole project with the fixes in place. Order matters and is correct.
- **`--frozen-lockfile` in CI** is the right call — CI should never silently resolve a different lockfile than `main`.
- **`actions/checkout@v4`, `actions/setup-node@v4`, `pnpm/action-setup@v4`** — all current major versions. `timeout-minutes: 5` matches the full-gate budget.
- **`VITE_TMDB_READ_TOKEN` from `secrets.*`** — the env-validator from PR #55 needs this in CI or the gate will fail at the format step (env.ts throws at module load when imported by main.tsx... actually wait, verify.sh doesn't import main.tsx). Correct, but flagging because someone reading might assume this is a global requirement.
- **`prepare: "husky"`** is the right hook for `core.hooksPath = .husky` to land on every fresh clone. Verified the field name is correct for both npm and pnpm.
- **`lint-staged` config** scoped to `*.{ts,tsx}` (eslint + prettier) and `*.{json,css,md}` (prettier only). Reasonable defaults; no surprises.

### Should look at (real, but not a blocker)

1. **`pnpm check-types` is still the broken pre-existing script flagged in PR #35.** `package.json` line 14: `"check-types": "tsc --noEmit"`. The root `tsconfig.json` has `"files": []` and project references, so `tsc --noEmit` (without `-b`) reads the root config, sees no files, and exits 0 — without actually type-checking anything. The gate step passes for the wrong reason: it does no work.

   This means the gate currently has a silent hole. A commit that breaks types will pass `pnpm check-types` because the script is a no-op, then fail at `pnpm build` (which runs `vite build` → type-checks as part of bundling). So the protection is still there, just at a different step and with a confusing failure surface.

   The fix is one line in `package.json`:
   ```diff
   -    "check-types": "tsc --noEmit",
   +    "check-types": "tsc -b --noEmit",
   ```
   Worth filing as a follow-up PR before this one merges, OR fold into this PR as a one-line addition. The PR's Quality Gate section would then be honest about every step actually checking something.

2. **`scripts/check-versions.sh` is slow by design.** It calls `pnpm view` once per dep (30+ in this repo), sequentially. Each call is a separate network round-trip to the npm registry. For a clean baseline, expect ~30–60s of just dep-check time, eating most of the 90s full-gate budget. Two improvements that would help:
   - **Batch**: `pnpm view --json pkg1 pkg2 ...` returns a single JSON document. One network call, parse the result.
   - **Parallel**: `xargs -P 8` over the dep list.
   
   Not blocking, but the PR claims a < 90s budget and this script alone can blow it on a slow registry day. Worth a `time bash scripts/check-versions.sh` before merging to see the real number.

3. **`pnpm view` has no timeout.** If the registry hangs (it happens), the gate hangs. Wrap the call:
   ```bash
   latest=$(timeout 10 pnpm view "$dep" version 2>/dev/null || echo '?')
   ```
   Or use `npm view` which is typically faster than pnpm's view command (pnpm hits the same registry but adds its own caching layer).

### Nits (not blocking)

4. **Trailing newlines missing** on `scripts/verify.sh`, `scripts/check-versions.sh`, `.husky/pre-commit`, `.husky/pre-push`. None of these are Prettier-checked (shell files aren't in the glob), so `pnpm format:check` passes — but the rest of the repo follows the "newline at EOF" convention from PR #51, and the inconsistency will bite when the next contributor edits one of these files in an editor that auto-strips trailing newlines and Prettier doesn't catch them.

5. **Step numbering is `[1/5]` even in `--quick` mode** which only runs 3 steps. Cosmetic, but a `case "$MODE"` to switch the labels to `[1/3]` for quick would be more honest about what's running.

6. **`check-versions.sh --gate` step is not wrapped in `step()`** in `verify.sh`. The other steps get the `▶` prefix; this one prints its raw output without the prefix. If it fails, the error message doesn't have the same shape as the other failures. Three-line fix.

7. **`check-versions.sh` distinguishes `DEPRECATED` from `has $latest`** — explicit intent in the PR body's "Reviewer Focus" #2. Worth a one-line comment in the script at the relevant `if` branch so the intent survives a future refactor.

8. **CI uses mutable action refs** (`actions/checkout@v4`, `actions/setup-node@v4`, `pnpm/action-setup@v4`). Pinning to commit SHAs is the supply-chain hardening play. `@v4` floats to whatever the latest v4.x tag points at. For a project whose whole point is "no surprises on `main`", this is the kind of footgun the gate is supposed to prevent. Low risk in practice; high risk in theory. Worth filing as a follow-up.

9. **No `concurrency:` block in `ci.yml`.** Concurrent PRs from the same branch can race on the shared `pnpm` cache. Standard hardening:
   ```yaml
   concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}
     cancel-in-progress: true
   ```

10. **No `permissions:` block in `ci.yml`.** GitHub Actions defaults to `permissions: write-all`. Best practice is to scope to the minimum:
    ```yaml
    permissions:
      contents: read
    ```

### Verdict

Approve. Item 1 is the one real concern — the gate has a silent type-check hole because the underlying `pnpm check-types` script is a no-op. The protection is still there (it surfaces at `pnpm build` instead), but the PR claims `pnpm check-types passes` which is technically true and semantically misleading. Folding the one-line `package.json` fix into this PR would close the loop. Everything else is polish.

This is the most important PR of the baseline. After it lands and the branch-protection rule is enabled (#27), `main` with a warning or a deprecated dep becomes structurally impossible — which is the project guide's loudest claim about how the work happens.

PR #67, opened 2026-08-21
feat(domain): model states, money, rating, and absence (#11)

3105jero:
# Pull Request

## Description

Issue #11 introduces the domain layer as the foundation of Cineteca — pure TypeScript with no framework dependencies. Every UI component that comes later (Explore, Home, Detail, MyCineteca) will consume these types.

Closes #11
Closes #19
Closes #20
Closes #21
Closes #22

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Chore / Maintenance (tooling)

## What Changed

- `src/domain/shared/no-data.ts` — `NoData<T>` tagged union + `absent()` / `present()` / `matchNoData()` (#22)
- `src/domain/movie/movie-state.ts` — `MovieState` (released / unreleased / unknown) + exhaustive `matchMovieState()` (#19)
- `src/domain/rating/rating-reliability.ts` — `RatingReliability` (consolidated / early / unknown) built from raw TMDB values via `ratingReliability(votes, average)` (#20)
- `src/domain/money/iso-4217.ts` — `ISO4217` branded currency code (rejects a plain `string`)
- `src/domain/money/money.ts` — `Money` interface + `money()` / `moneyFromMajor()` / `moneyFromTMDB()` with minor-unit exponent table (#21)
- `src/domain/money/format-money.ts` — `formatMoney()` delegating to `Intl.NumberFormat`; locale defaults to `navigator.language` with `en-US` fallback (#21)
- `vitest.config.ts` — enabled coverage thresholds (100% on `src/domain/**`, 80% global)
- `eslint.config.js` — extended ignore glob to `*.config.ts` so `vitest.config.ts` no longer trips the lint-staged `--max-warnings 0` check
- `package.json` — added `--no-warn-ignored` to the lint-staged eslint command
- `.gitignore` — ignore the personal `Cineteca.md` / `Cinética BaseLine.md` copies kept at the workspace root
- `src/domain/shared/.gitkeep` — removed (the directory now has real files)

84 new tests, all green. 100% coverage in `src/domain/`.

## Affected Layers

- [x] `domain/` — pure TypeScript, no framework imports

## Screenshots / Recordings

N/A — no UI.

## How to Test

1. Pull the branch: `git fetch && git checkout feature/domain-model-states-money-rating`
2. Run `pnpm install`
3. Run `pnpm test` — expect 84 passing, 100% in `domain/`
4. Run `bash scripts/verify.sh --full` — expect green in ~65s

## Tests

[...]

Issue #12, opened 2026-08-20
[Feature]: TMDB API client + validation


3105jero:
## Description
Implement the HTTP client, the API module per endpoint, and the validation layer.

## Tasks
- [ ] Create one Axios instance in `src/infrastructure/http/` with interceptors
- [ ] Read the token from `env.VITE_TMDB_READ_TOKEN`
- [ ] Translate TMDB error codes (e.g. 34 for "not found") to domain errors **once, at the edge**
- [ ] Respect `429` rate limits — wait the time the server indicates
- [ ] Implement one module per endpoint in `src/infrastructure/api/`:
  - `configuration` — image base + sizes, cached forever
  - `genres` — movie genres for filters
  - `discover` — discovery with filters (genre, year, min rating, min votes, sort)
  - `search` — text search
  - `movie` — detail with optional append (cast + trailers in one call)
  - `recommendations` — for movie detail's lower section
  - `trending` — for the home screen
- [ ] Every response is validated with Zod before it leaves the infrastructure layer
- [ ] Cover the API surface with MSW handlers in `src/test/msw/`

## Acceptance Criteria
- [ ] No `axios` import outside `src/infrastructure/http/`
- [ ] Every API function returns a typed domain object, not raw TMDB JSON
- [ ] A broken response in tests produces a clear, localized error
- [ ] Breaking a field in the simulation shows the error reaches the UI

## Reference
[Cineteca.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cineteca.md) — "El contrato con la API que consumen".

## Dependencies
- Depends on #11 Domain modeling

Issue #71, opened 2026-08-21
docs(contributing): enforce one Conventional-Commit type per PR (split tooling out of feature PRs)


SrLampi1001:
## Motivation

[PR #67 review](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/pull/67#pullrequestreview-4994061831), "Should look at" item #5:

> **Tooling changes folded into a domain PR.** `package.json` (lint-staged flag), `eslint.config.js` (ignore glob extension), `vitest.config.ts` (coverage thresholds), and `.gitignore` (personal-guide copies) are all unrelated to the four sub-issues. The contributing convention says each PR is one Conventional-Commit type — these are `chore(tooling)`/`chore(test)`, not `feat(domain)`. Not blocking, but worth keeping in mind for PR #70+; likely worth an ADR or contributing-guide note.

PR #67's author flags the folding in their own "Reviewer Focus" section — *"Tooling fixes folded in. The eslint ignore extension and `--no-warn-ignored` flag are not domain work but were needed for the pre-commit hook to accept the new `vitest.config.ts`."* That makes the precedent explicit: tooling gets folded when its friction blocks the feature. The next PR will face the same friction and reach for the same shortcut.

The current contributing docs (`CONTRIBUTING.md` + `.github/PULL_REQUEST_TEMPLATE.md`) list every gate step as a checkbox but don't gate on *scope*. The two-line rule that the previous session's PRD mentions — *"Each PR is one Conventional-Commit type"* — is implicit at best. Worth making it explicit before PR #70 inherits the same pattern.

## Proposed Approach

Two edits, both docs-only.

### 1. Add a "Scope" section to `CONTRIBUTING.md`

Right after the existing "Pull request" section, add a one-paragraph rule:

> **One commit type per PR.** A PR titled `feat(domain): …` should change domain code only. Lint-staged flags, eslint ignore globs, vitest coverage thresholds, dependency bumps, and `.gitignore` edits belong in their own `chore(...)` PR — even when the original feature PR can't land without them. Fold-in is allowed *only* when the friction is genuinely blocking and can't be sidestepped, and even then the PR title carries the dominant type. Reference the fold-in with a sentence in the body so reviewers see it.

A short follow-up paragraph gives a one-line test: *"If your git diff shows two layers of the architecture changing for different reasons, that's two PRs."*

### 2. Mirror the rule in `.github/PULL_REQUEST_TEMPLATE.md`

[...]

PR #72, opened 2026-08-21
feat(infra): add TMDB HTTP client with interceptors and error translation

### LINKED ISSUE #23: [Feature]: HTTP client + interceptors
## Description
Build the single Axios instance with interceptors. This is the only place that knows axios.

## Tasks
- [ ] Create `src/infrastructure/http/client.ts` with a single Axios instance
- [ ] Inject `Authorization: Bearer ${env.VITE_TMDB_READ_TOKEN}` automatically
- [ ] Translate TMDB error codes (their numeric codes don't match HTTP) to a domain error type
- [ ] On `429` rate-limit, wait the time the server indicates and retry once
- [ ] Export typed errors from the boundary

## Acceptance Criteria
- [ ] No `axios` import exists outside `src/infrastructure/http/`
- [ ] MSW intercepts the test calls — no real network hits
- [ ] A simulated 429 from TMDB triggers the wait

## Parent
#12 — TMDB API client + validation

3105jero:
# Pull Request

## Description

Issue #23 builds the single Axios instance the rest of the codebase talks to. The dependency rule (`eslint.config.js`) bans `axios` imports outside `src/infrastructure/http/**`; this PR fills that directory with the only piece of code allowed to know axios.

Closes #23

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## What Changed

- `src/infrastructure/http/errors.ts` — `TmdbError` discriminated union + `TmdbHttpError` class + `translateError()` that turns a raw Axios failure into a typed error carrying `kind: 'notFound' | 'invalidPage' | 'invalidApiKey' | 'rateLimited' | 'serverError' | 'networkError' | 'unknown'`
- `src/infrastructure/http/client.ts` — singleton axios instance with two interceptors:
  - **Request**: injects `Authorization: Bearer ${env.VITE_TMDB_READ_TOKEN}` on every call
  - **Response**: retries once on 429 with the server-indicated `Retry-After` (defaults to 1s); translates every other failure into `TmdbHttpError`
- `src/infrastructure/http/client.ts` — exports `tmdbHttpClient` (typed `get` / `post` / `put` / `delete` that return response bodies, not axios envelopes) and re-exports `TmdbHttpError`
- `src/infrastructure/http/client.spec.ts` — 13 tests via MSW: auth header, every translation branch, 429 retry-then-succeed, 429 retry-then-fail, missing-Retry-After default
- `src/infrastructure/http/.gitkeep` — removed (directory now has real files)

## Affected Layers

- No `domain/` or `presentation/` changes.

## Screenshots / Recordings

N/A — no UI.

## How to Test

1. Pull the branch: `git fetch && git checkout feature/infra-http-client-interceptors`
2. Run `pnpm install`
3. Run `pnpm test` — expect 97 passing, http module ~98% lines / 89% branches
4. Run `bash scripts/verify.sh --full` — expect green in ~70s

Spot-checks worth running by hand (in a Vitest REPL or temporary test):

```ts
// Happy path
const { tmdbHttpClient } = await import('./src/infrastructure/http/client');
const data = await tmdbHttpClient.get('/3/configuration'); // throws if env not set

// Error path
try {
  await tmdbHttpClient.get('/3/missing');
} catch (e) {
  // e.detail.kind === 'notFound'
}
```

## Tests

- [x] I added unit tests for the new logic
- [ ] I updated existing tests that no longer apply (none affected)
- [x] Coverage has not decreased (global 98.77% statements / 91.66% branches; http module 98.5% lines / 88.5% branches)

## Quality Gate

- [x] `pnpm build` succeeds

Local run: `bash scripts/verify.sh --full` → `✔ FULL GATE GREEN in 69s`.

## Checklist

- [x] My code follows the project's architecture rules (axios is imported only in `infrastructure/http/`; the linter would reject any other location)

## Deployment Notes

None. The client is only used by code that doesn't exist yet (the resource adapters under `src/infrastructure/api/<resource>/`); nothing in the existing app shell imports it. The PR is a foundation for #12.

## Reviewer Focus

A few decisions worth a second look:

- **`TmdbHttpError` as a class, not a tagged value.** The rest of the app needs to `throw` it, and `unknown` rejection patterns are easier to debug when the error has a `.name`. The discriminated `detail.kind` is what consumers actually pattern-match on.
- **`__tmdbRetried` module augmentation.** A `boolean` flag on `InternalAxiosRequestConfig` via `declare module 'axios'`. The alternative was a `WeakMap<InternalAxiosRequestConfig, boolean>`, which is heavier and harder to debug. This is the lightest non-leaky option.
- **Translation table is partial.** TMDB's documented codes (22, 34, 7/10/30/31) are mapped explicitly; everything else falls through to HTTP-status-based kinds (5xx → `serverError`, other 4xx → `unknown`). The comment in `translateError` flags this as a deliberate scope choice — extending the table is a one-PR change when a feature needs it.
- **`Retry-After` default of 1 second, and `0` is valid.** `Number.parseInt('0', 10) === 0`, which `> 0` rejected; I switched the condition to `>= 0` so the server can ask for an immediate retry. Missing/non-numeric/negative headers fall back to 1 second.
- **Module-level singleton.** The issue said "single Axios instance" — I went with `const tmdbHttpInstance = createTmdbHttpInstance()`. Tests get a fresh instance per test via `vi.resetModules()` + dynamic import; this is the same pattern `env.spec.ts` uses.
- **No tooling or governance changes folded in this time.** Scoped strictly to `src/infrastructure/http/` per the reviewer feedback on PR #67.

---

### DIFF
--- a/src/infrastructure/http/client.spec.ts
+++ b/src/infrastructure/http/client.spec.ts
@@ -1,0 +1,248 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { http, HttpResponse } from 'msw';
+import { server } from '@/test/msw/server';
+
+const BASE = 'https://api.tmdb.test';
+const TOKEN = 'a'.repeat(64);
+
+describe('infrastructure/http/client', () => {
+  beforeEach(() => {
+    // resetModules must run before the import below so the client module
+    // re-evaluates with the freshly-stubbed env on every test.
+    vi.resetModules();
+    vi.unstubAllEnvs();
+    vi.stubEnv('VITE_TMDB_READ_TOKEN', TOKEN);
+    vi.stubEnv('VITE_TMDB_API_BASE', BASE);
+    server.resetHandlers();
+  });
+
+  afterEach(() => {
+    server.resetHandlers();
+    vi.useRealTimers();
+  });
+
+  describe('request interceptor', () => {
+    it('returns the JSON body on 200', async () => {
+      server.use(http.get(`${BASE}/3/ping`, () => HttpResponse.json({ ok: true })));
+
+      const { tmdbHttpClient } = await import('./client');
+      const result = await tmdbHttpClient.get<{ ok: boolean }>('/3/ping');
+      expect(result).toEqual({ ok: true });
+    });
+
+    it('attaches the Bearer token from env on every request', async () => {
+      let received: string | null = null;
+      server.use(
+        http.get(`${BASE}/3/whoami`, ({ request }) => {
+          received = request.headers.get('authorization');
+          return HttpResponse.json({ ok: true });
+        }),
+      );
+
+      const { tmdbHttpClient } = await import('./client');
+      await tmdbHttpClient.get('/3/whoami');
+      expect(received).toBe(`Bearer ${TOKEN}`);
+    });
+
+    it('sends POST, PUT and DELETE bodies through to the server', async () => {
+      const seen: { method: string; body: string }[] = [];
+      server.use(
+        http.post(`${BASE}/3/echo`, ({ request }) =>
+          request.text().then((body) => {
+            seen.push({ method: 'POST', body });
+            return HttpResponse.json({ ok: true });
+          }),
+        ),
+        http.put(`${BASE}/3/echo`, ({ request }) =>
+          request.text().then((body) => {
+            seen.push({ method: 'PUT', body });
+            return HttpResponse.json({ ok: true });
+          }),
+        ),
+        http.delete(`${BASE}/3/echo`, () => {
+          seen.push({ method: 'DELETE', body: '' });
+          return HttpResponse.json({ ok: true });
+        }),
+      );
+
+      const { tmdbHttpClient } = await import('./client');
+      await tmdbHttpClient.post('/3/echo', { a: 1 });
+      await tmdbHttpClient.put('/3/echo', { a: 2 });
+      await tmdbHttpClient.delete('/3/echo');
+
+      expect(seen.map((s) => s.method)).toEqual(['POST', 'PUT', 'DELETE']);
+      expect(seen[0]?.body).toBe(JSON.stringify({ a: 1 }));
+      expect(seen[1]?.body).toBe(JSON.stringify({ a: 2 }));
+    });
+  });
+
+  describe('error translation', () => {
+    it('translates TMDB code 34 + HTTP 404 into notFound', async () => {
+      server.use(
+        http.get(`${BASE}/3/movie/999`, () =>
+          HttpResponse.json(
+            { success: false, status_code: 34, status_message: 'Not found.' },
+            { status: 404 },
+          ),
+        ),
+      );
+
+      const { tmdbHttpClient } = await import('./client');
+      const promise = tmdbHttpClient.get('/3/movie/999');
+      await expect(promise).rejects.toMatchObject({
+        name: 'TmdbHttpError',
+        detail: { kind: 'notFound', message: 'Not found.' },
+      });
+    });
+
+    it('translates TMDB code 22 + HTTP 400 into invalidPage', async () => {
+      server.use(
+        http.get(`${BASE}/3/discover/movie`, () =>
+          HttpResponse.json(
+            { success: false, status_code: 22, status_message: 'Invalid page.' },
+            { status: 400 },
+          ),
+        ),
+      );
+
+      const { tmdbHttpClient } = await import('./client');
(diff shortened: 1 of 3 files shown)


SrLampi1001 (approved):
**Verdict: APPROVE** — first-pass review. The PR delivers #23 cleanly: axios is the only framework that knows itself, every public surface returns `TmdbHttpError` rather than raw axios, and MSW covers all the translation branches plus the retry path. The gate is claimed green (`✔ FULL GATE GREEN in 69s`) and the diff is scoped strictly to `src/infrastructure/http/` per the contributor guideline learned from PR #67.

---

### Blockers

None. Architecture compliance is real, not paper — `axios` is imported in exactly one file (`src/infrastructure/http/client.ts:32`), and the linter's catch-all ban (`eslint.config.js:51-67`'s `paths: [{ name: 'axios' }]` for `src/**` with `ignores: ['src/infrastructure/http/**']`) confirms no other layer can reach for it.

### Should look at (filed as follow-up issues below)

1. **Test gap on `invalidApiKey` codes.** `src/infrastructure/http/errors.ts:127` maps TMDB codes `7 / 10 / 30 / 31` to `invalidApiKey`, but only code 7 is asserted in `client.spec.ts`. If a future refactor drops one of the other three from the table, no test fails. Extracting `const INVALID_API_KEY_TMDB_CODES = [7, 10, 30, 31] as const` and table-driving the test, or adding three more micro-cases, closes the gap.

2. **`Retry-After` as an HTTP-date is unhandled.** `errors.ts:115` and `client.ts:78` both parse `Retry-After` with `Number.parseInt(retryAfterHeader, 10)` — the seconds form only. RFC 9110 §10.2.3 also permits an HTTP-date. A spec-compliant server could send `Retry-After: Wed, 21 Oct 2015 07:28:00 GMT` and the client would silently fall back to the 1-second default. Worth tracking as a `feat(infra):` and putting the parser in `errors.ts` so both paths share it.

3. **`tmdbHttpClient` methods have no JSDoc.** `client.ts:88-101` (`get` / `post` / `put` / `delete`) carries only an inline comment about the wrapping purpose. Each method should declare its parameter semantics (`config` is the standard axios `AxiosRequestConfig`; `body` is JSON-serialized by axios) and the resolved type. Consumers today have to read the source to learn that the resolved value is `response.data`, not the full axios envelope — which is the entire point of the wrapper.

4. **Default-retry "1 second" is duplicated.** `client.ts:78` and `errors.ts:115` each have a separate `parsed >= 0 ? parsed : 1`. The two defaults are independent (one gates the `sleep`, the other populates `rateLimited.retryAfterSeconds`) and that's a footgun for whoever next edits the policy. A `DEFAULT_RETRY_AFTER_SECONDS` constant exported from `errors.ts` centralizes it.

### Nits

- `client.ts:69` — `config.headers.set('Authorization', ...)` is the axios v1 `AxiosHeaders.set` API. The package is pinned at `axios@^1.19.0` (`package.json`), so this is correct; a one-line comment next to the pin would save the next reader a docs check.
- `errors.ts:75-80` — `messageFor` reads `error.message` first. If axios ever returns a useful multi-line message (it sometimes does for CORS / DNS chains), the message ends up as the `status_message`. Preferring `body.status_message` when present would keep TMDB-sourced messages on top.
- `client.ts:67` — `const headers = response.headers as Record<string, string | undefined>` casts through `unknown` implicitly. axios v1's `AxiosResponseHeaders` has a typed `.get()` accessor; `response.headers.get('retry-after')` drops the cast.

### What's good

- `TmdbError` as a `readonly` discriminated union is the right shape for the `@typescript-eslint/switch-exhaustiveness-check` rule (`eslint.config.js:35`). Consumers that fall through a `switch (e.detail.kind)` will fail to type-check — a feature, not a bug.
- The retry condition is `>= 0` (in both `client.ts:78` and `errors.ts:115`), so the server can ask for an immediate retry. Correct reading of RFC 9110 §10.2.3.
- PR scope is narrow. The contributor learned from PR #67's folding-in feedback: "Reviewer Focus" promises *"No tooling or governance changes folded in this time"* and the diff honors that (4 files, 521 insertions, all in one directory). Exactly the discipline #71 will codify.
- `translateError`'s duck-typed input (not `AxiosError`) makes the translation logic unit-testable in isolation. That's what enables the direct `translateError({})` test on `client.spec.ts:169` — a nice property to keep.
- 13 tests cover the interceptors, every translation branch, retry-then-succeed, retry-then-fail, and the missing-header default. The spec file is the contract.

---

**Verdict: APPROVE.** Ready for squash-merge into `main` per the project's workflow (`feat(infra): subject (#72)`). The four follow-ups above will be filed as issues before the merge body references them.

Issue #24, opened 2026-08-20
[Feature]: TMDB endpoint modules + Zod schemas


3105jero:
## Description
Implement the data-fetching modules for the five core endpoints. Every response is validated with Zod before it leaves this layer.

## Tasks
- [ ] `configuration` — image base + sizes, cached forever
- [ ] `genres` — list for filter dropdowns
- [ ] `discover` — discovery with filters (genre, year, minRating, minVotes, sort, page)
- [ ] `search` — text search with debounce
- [ ] `movie` — detail with optional `append_to_response` (cast + trailers in one call)
- [ ] `trending` — for the home screen
- [ ] Zod schema for every response
- [ ] MSW handlers for all endpoints using real API response shapes

## Acceptance Criteria
- [ ] Every API function returns a typed domain object, never raw TMDB JSON
- [ ] Breaking a field in MSW produces a localized error
- [ ] Pagination respects the 500-page hard cap
- [ ] No `any` in any API module

## Parent
#12 — TMDB API client + validation

Issue #19, opened 2026-08-20
[Feature]: Movie state tagged union


3105jero:
## Description
Model the movie lifecycle as a discriminated union. Each branch carries exactly the data it has.

## Tasks
- [ ] Define `MovieState` as a tagged union: `Released`, `Unreleased`, `Unknown`
- [ ] Each branch carries only the data it has — no nullable shared fields
- [ ] Build exhaustive `switch` over the union (covered by `switch-exhaustiveness-check`)
- [ ] Each state has its own label and visual treatment in the UI

## Acceptance Criteria
- [ ] The compiler enforces exhaustiveness in every consumer
- [ ] Unreleased movies show a text label, not just a gray border
- [ ] Tests cover all branches

## Reference
[Cineteca.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cineteca.md) — "El dominio: estados, dinero y ausencias".

## Parent
#11 — Domain modeling (states, money, rating)

PR #77, opened 2026-08-21
feat(infra): add per-endpoint TMDB API modules with Zod validation

### LINKED ISSUE #12: [Feature]: TMDB API client + validation

SebastianT2006:
Seven endpoints, one module each, each typed at the boundary:

- **configuration** — image base + sizes, cached forever via a lazy in-flight promise so the network is hit once per app lifetime.
- **genres** — movie genres for filters.
- **discover** — paginated discovery with filters (genre, year, min rating, min votes, sort, page, include_adult).
- **search** — paginated text search with query, page, include_adult, primary_release_year.
- **movie** — single-movie detail; `appendCastAndTrailers` flips on `append_to_response=credits,videos` so cast and trailers come back in the same round-trip.
- **recommendations** — paginated recommendations for a movie id.
- **trending** — paginated trending for day or week.

Every response is parsed by a Zod schema before it leaves infrastructure; a broken field surfaces a `ZodError`, never a silent wrong value. TMDB's raw `poster_path:null` / `release_date:''` become explicit `NoData` or `unknown MovieState` at the edge, so the domain stays free of TMDB.

Domain additions (composed from existing primitives, no axios/React imports): `MovieSummary`, `MovieDetail`, `CastMember`, `Trailer`, `Genre`, `PaginatedList`, `AppConfiguration`, `ImageConfiguration`.

MSW handlers in `src/test/msw/tmdb-handlers.ts` cover the seven endpoints with realistic fixtures — available to any future integration test, not used by these per-endpoint specs (each one owns its own handler for shape variations).

Closes #12

### DIFF
--- a/src/domain/configuration/app-configuration.spec.ts
+++ b/src/domain/configuration/app-configuration.spec.ts
@@ -1,0 +1,21 @@
+import { describe, expect, it } from 'vitest';
+import type { AppConfiguration } from './app-configuration';
+
+describe('domain/configuration/app-configuration', () => {
+  it('wraps the image configuration and exposes change_keys', () => {
+    const config: AppConfiguration = {
+      images: {
+        baseUrl: 'http://image.tmdb.org/t/p',
+        secureBaseUrl: 'https://image.tmdb.org/t/p',
+        posterSizes: ['w185'],
+        backdropSizes: ['w1280'],
+        profileSizes: ['w185'],
+        stillSizes: ['w300'],
+        logoSizes: ['w300'],
+      },
+      changeKeys: ['adult', 'air_date', 'also_known_as'],
+    };
+    expect(config.images.secureBaseUrl).toBe('https://image.tmdb.org/t/p');
+    expect(config.changeKeys).toContain('adult');
+  });
+});
--- a/src/domain/configuration/app-configuration.ts
+++ b/src/domain/configuration/app-configuration.ts
@@ -1,0 +1,16 @@
+import { type ImageConfiguration } from '@/domain/configuration/image-configuration';
+
+/**
+ * AppConfiguration — the TMDB `/configuration` response, normalized.
+ *
+ * Today only `images` is used; tomorrow a future feature may need
+ * `change_keys` (e.g. the kinds of fields TMDB allows you to append).
+ * Keeping the full shape now means a future change is purely
+ * additive on the consumer side.
+ *
+ * @see Cineteca.md — "Base y tamaños de imagen".
+ */
+export interface AppConfiguration {
+  readonly images: ImageConfiguration;
+  readonly changeKeys: readonly string[];
+}
--- a/src/domain/configuration/image-configuration.spec.ts
+++ b/src/domain/configuration/image-configuration.spec.ts
@@ -1,0 +1,19 @@
+import { describe, expect, it } from 'vitest';
+import type { ImageConfiguration } from './image-configuration';
+
+describe('domain/configuration/image-configuration', () => {
+  it('carries the secure base and the per-type size lists', () => {
+    const config: ImageConfiguration = {
+      baseUrl: 'http://image.tmdb.org/t/p',
+      secureBaseUrl: 'https://image.tmdb.org/t/p',
+      posterSizes: ['w92', 'w154', 'w185', 'w342', 'w500', 'w780', 'original'],
+      backdropSizes: ['w300', 'w780', 'w1280', 'original'],
+      profileSizes: ['w45', 'w185', 'h632', 'original'],
+      stillSizes: ['w92', 'w185', 'w300', 'original'],
+      logoSizes: ['w45', 'w92', 'w154', 'w185', 'w300', 'w500', 'original'],
+    };
+    expect(config.secureBaseUrl).toBe('https://image.tmdb.org/t/p');
+    expect(config.posterSizes).toContain('w500');
+    expect(config.posterSizes).toContain('original');
+  });
+});
--- a/src/domain/configuration/image-configuration.ts
+++ b/src/domain/configuration/image-configuration.ts
@@ -1,0 +1,27 @@
+/**
+ * ImageConfiguration — the size catalogue for image URLs.
+ *
+ * TMDB's `/configuration` endpoint hands back a list of sizes for
+ * each image type (poster, backdrop, profile, still, logo). Every
+ * application URL is built as
+ *   `${baseUrl}/${size}${path}`
+ * where `path` is what the per-resource endpoints return (e.g.
+ * `/abc.jpg`).
+ *
+ * Keeping this shape in the domain means the UI can build URLs
+ * without depending on TMDB's response layout. The
+ * `secureBaseUrl` is what production uses; `baseUrl` is here only
+ * to keep the contract round-trippable when reading the cached
+ * response back from storage in a future issue.
+ *
+ * @see Cineteca.md — "Base y tamaños de imagen".
+ */
+export interface ImageConfiguration {
+  readonly baseUrl: string;
+  readonly secureBaseUrl: string;
+  readonly posterSizes: readonly string[];
+  readonly backdropSizes: readonly string[];
+  readonly profileSizes: readonly string[];
+  readonly stillSizes: readonly string[];
+  readonly logoSizes: readonly string[];
+}
--- a/src/domain/movie/cast.spec.ts
+++ b/src/domain/movie/cast.spec.ts
@@ -1,0 +1,24 @@
+import { absent, present } from '@/domain/shared/no-data';
+import { describe, expect, it } from 'vitest';
+import type { CastMember } from './cast';
+
+describe('domain/movie/cast', () => {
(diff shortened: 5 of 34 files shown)


3105jero (approved):
This is the data-layer foundation for the entire app, and it's exceptional code. 38 files, ~2900 additions, all architected cleanly. The HTTP client alone (retry-once-on-429 with the `__tmdbRetried` flag, `TmdbHttpError` discriminated kind, no `axios` leaking past `src/infrastructure/http/**`) is the kind of work that pays for itself the first time a real-world response surprises us. Approving with one literal spec gap and a handful of nits.

### What works

- **HTTP client is bullet-proof.** Three things stand out: (1) the `__tmdbRetried` module-augmentation flag prevents an infinite retry loop on a pathological 429, (2) `Retry-After` parsing with the `parsed >= 0` guard so a `0` means "retry immediately" rather than falling back to 1, (3) the surface returns `T` directly (`tmdbHttpClient.get<T>(url)` returns `T`, not `{ data, status, headers, … }`) so callers can't accidentally reach into a raw Axios shape. The `REQUEST_TIMEOUT_MS = 10_000` default is a sane starting point.
- **`TmdbHttpError` is the right shape.** Tagged union (`rateLimited` / `notFound` / `invalidPage` / `invalidApiKey` / `serverError` / `networkError` / `unknown`) instead of an error-class hierarchy. The mapping from TMDB's numeric body codes (22, 34, 7/10/30/31) to kinds is documented in the file header and verified by the spec.
- **Architecture compliance is clean.** Verified at the import level: `src/domain/**` only imports `vitest` (in specs) and other `src/domain/**` modules. No `axios`, no `react`, no `react-*`, no `@tanstack/*`, no `react-hook-form`. The ESLint `no-restricted-imports` rule from PR #43 passes every domain file by construction. `src/infrastructure/api/**` imports only `zod`, `tmdbHttpClient`, and `src/domain/**` types — never raw `axios`. The dependency rule holds.
- **Raw → domain mappers are pure and total.** `classifyRelease(rawReleaseDate, now)` takes `now` as a parameter (not `Date.now()`) so tests can pin time deterministically. `ratingFromVotes(voteCount, voteAverage)` translates `vote_count === 0` to `unknown` so the type carries "we don't have a rating" rather than a meaningless "0 of 10".
- **Null/`''` translation at the boundary, not in the domain.** TMDB's `poster_path: null` becomes `NoData.absent` in `MovieSummary.posterPath`. Empty `release_date` becomes `unknownMovieState`. The domain stays free of TMDB-shaped values — verified by reading `movie-summary.ts` and its mapper in `_shared.ts`.
- **`configuration` cached forever via a single in-flight promise.** `cached ??= fetchAndParse()` is the right shape: concurrent callers see the same promise (no duplicate network requests), and the cache lives for the module's lifetime (resets on HMR, per the file comment). `__resetConfigurationCacheForTests` is properly scoped to a clear name with the test-only purpose documented.
- **`_shared.ts` naming convention** (`tmdb*Schema` for raw Zod, `to*` for the pure mapper) is documented in the file header. The underscore prefix tells readers "infrastructure-only; don't import from outside".
- **MSW handlers are designed to be extendable per-test.** The header documents that each per-endpoint spec owns its own handler for shape variations; the bundled `tmdb-handlers.ts` is for tests that just want "any valid TMDB response". The handlers are realistic snapshots with `http.get` factories that compose with `server.use(...)`.
- **TMDB body codes get documented coverage for the codes Cineteca cares about** (22 = invalid page, 34 = not found, 7/10/30/31 = auth cluster). The PR header acknowledges the table is intentionally partial. That's the right call — a speculative full table is a maintenance burden for codes that don't drive any feature.
- **Format gate passes.** `pnpm prettier --check .` returns "All matched files use Prettier code style!" against the project's `.prettierrc.json` and `prettier@3.9.6`.

### Real concern (one spec acceptance criterion not met)

1. **The 500-page hard cap is not enforced anywhere in this PR.** Issue #24 acceptance criteria: *"Pagination respects the 500-page hard cap"*. Today, every paginated endpoint (`discover`, `search`, `trending`, `recommendations`) passes `filters.page` through to TMDB unmodified. `paginated.ts` describes the shape but its comment says *"downstream consumers enforce it"* — i.e., the UI layer — and there is no UI layer yet to enforce it.

   The trivial fix in `_shared.ts` is one helper that all four endpoints call:
   ```ts
   export const MAX_PAGE = 500;
   export function clampPage(p: number | undefined): number | undefined {
     return p === undefined ? undefined : Math.min(Math.max(1, Math.floor(p)), MAX_PAGE);
   }
   ```
   Then each API module's `toQueryParams` does `page: clampPage(filters.page)` (or returns `undefined` when the call didn't pass a page at all). Three lines per file, one helper.

   Defense-in-depth argument: the API layer is the last place that has TMDB-shaped context. If the UI forgets the clamp, the API should still catch it. Otherwise the gate stays open on every consumer that doesn't replicate the rule. Worth folding into this PR — it's a literal spec item and the fix is mechanical.

2. **`includeAdult` is exposed in three modules (`discover`, `search`, `trending`) without an app-level policy.** TMDB asks for it explicitly in their API guide, but the Cineteca project guide doesn't say whether the app shows adult content. The current shape lets any future caller flip the flag. Either remove `includeAdult` from the params interface until a feature decides to use it, or add a one-line comment in each module: *"Cineteca currently surfaces only non-adult content; this is here so the feature can flip the policy without an API change"*. The first is cleaner.

3. **TMDB body-code mapping is explicitly partial (ack'd in the file header).** When a future feature hits an unmapped TMDB code (e.g. `account_id` issues for session endpoints, which TMDB will surface later if any feature needs them), it silently falls into `unknown`. Worth a `// See Cineteca.md §Errors for the documented TMDB codes. Extend when a feature needs.` comment near the `code` switch so the next reader knows where to look.

### Nits (not blocking)

4. **`paginated.ts` interface uses `readonly T[]`** for the results array. The Zod parsing produces mutable arrays, so this works only because `results` is assigned from `.map((m) => toMovieSummary(m, now))` — a fresh array. Defensive: the Zod schema could include `.readonly()` on the results array so the immutability is enforced at the boundary, not asserted at the type level. Optional.

5. **`movie-summary.ts` exports `classifyRelease` and `ratingFromVotes` as named functions** alongside the interface. Some teams prefer these as `static` methods on the interface or as a single `MovieSummary.from(...)` factory. The current style is more functional and easier to test in isolation. Keep as-is; flagging only because the codebase will face the same decision when `MovieDetail` gets similar factories.

6. **`recommendMovies` (PR body says `recommendations`) — naming mismatch between body and method name.** The endpoint in the file is `recommendMovies`, the PR body and issue spec say "recommendations". Pick one. (`recommendMovies` is the singular verb for one call; `recommendations` reads more naturally for a paginated list of recommendations. The body/spec wins for the user-facing naming.)

7. **`__tmdbRetried?: boolean` is augmented on `InternalAxiosRequestConfig`** but never reset between production calls. In practice, Axios creates a new config object per call (or shallow-copies), so the flag doesn't leak across requests. But if Axios ever pools config objects, a freshly-issued request could see `__tmdbRetried: true` from a previous request and refuse to retry. Defensive cleanup at the top of the response interceptor would harden this — at the cost of one comment explaining why.

### Verdict

Approve. Item 1 is the literal spec gap; fix in a follow-up commit (one helper, four call sites — 15 minutes of work). Items 2-7 are polish.

This is the data-layer foundation every future feature PR will sit on top of. The TMDB response shapes, the error kinds, the `NoData` and `MovieState` translations, and the HTTP client's retry behavior — all defined here, all tested here, all architected correctly. The next PR (`#11` rating reliability, `#19-22` domain types) can land on top of this without changing any of the contracts introduced here.

Issue #16, opened 2026-08-20
[Feature]: Local lists (create, edit, validate)


3105jero:
## Description
Users can create themed lists of movies (e.g. "90s noir"). Lists support create, edit, delete, and add/remove movies.

## Tasks
- [ ] Create List with React Hook Form + Zod schema (one schema validates and types)
- [ ] Validation rules: name 1-50 chars, description optional, max 20 chars
- [ ] Each validation rule has its own message, never generic "can't do"
- [ ] Submit blocked while in-flight; explicit success state
- [ ] Accessible: errors announced, focus moves to first invalid field
- [ ] Add/remove movies from a list with optimistic updates
- [ ] List detail page route

## Acceptance Criteria
- [ ] One Zod schema validates and types the form
- [ ] Each rule has its own Spanish message
- [ ] Tested by accessible role/name — no `data-testid`
- [ ] Errors are announced to screen readers
- [ ] Focus moves to the first invalid field on submit

## Reference
[Cineteca.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cineteca.md) — "Formularios".

## Dependencies
- Depends on #15 My Cineteca

Issue #15, opened 2026-08-20
[Feature]: My Cineteca (local library + optimistic mutations)


3105jero:
## Description
The user's library lives in localStorage. Save and remove with optimistic updates, storage validated **on read**.

## Tasks
- [ ] Create `src/infrastructure/storage/` adapter for browser storage
- [ ] Validate data **on read** with Zod — corrupted data is discarded gracefully
- [ ] Save / remove with optimistic mutation: instant UI, revert on failure
- [ ] Invalidate cache across views when a movie is saved/removed
- [ ] Back button after a save shows the **new** state, not the old one
- [ ] Empty state: "Your cineteca is empty" with a link to explore
- [ ] Tests prove: corrupted storage doesn't crash the app

## Rules
- **Storage is an untrusted edge** — same as the network
- **No silent data loss** — corruption is logged, not swallowed

## Acceptance Criteria
- [ ] Save shows instantly in the UI
- [ ] Simulated storage failure reverts the UI
- [ ] Corrupted data in storage is discarded, not crash
- [ ] Cross-cache invalidation verified (e.g. detail view → library view)

## Reference
[Cineteca.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cineteca.md) — "El estado de la vista vive en la URL; el estado del servidor vive en la caché."

## Dependencies
- Depends on #14 Home + Search + Detail

Issue #14, opened 2026-08-20
[Feature]: Home, Search, and Movie Detail screens


3105jero:
## Description
Home with weekly trending; search with debounce; movie detail with cast and trailers.

### Home
- [ ] Fetch trending of the week
- [ ] Entry point to explore

### Search
- [ ] Free text search with debounce — 10 keystrokes trigger **one** request, not ten
- [ ] Empty state for "no results"
- [ ] Cancel in-flight requests on new input

### Movie Detail
- [ ] Route `/movie/:id` — shareable URL
- [ ] Use the **append parameter** to fetch detail + cast + trailers in one call
- [ ] Recommendations section at the bottom
- [ ] All four states (loading, error, empty, success)
- [ ] Accessible: `<h1>` is the movie title
- [ ] Document title and focus move on route change
- [ ] Missing synopsis: offer English version with a clear notice

## Acceptance Criteria
- [ ] Search debounce verified in DevTools Network — typing fast triggers 1 request
- [ ] Movie detail URL is shareable and reloadable
- [ ] Unreleased movie shows "unreleased" status (not by color alone)
- [ ] Missing budget/revenue shows "no data", never "$0"
- [ ] Missing votes shows "no ratings", never `0.0`

## Reference
[Cineteca.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cineteca.md) — "Las pantallas y los cuatro estados".

## Dependencies
- Depends on #13 Explore screen

PR #78, opened 2026-08-21
feat(presentation): Explore screen with URL filters and four states

### LINKED ISSUE #13: [Feature]: Explore screen with filters and four states
## Description
First vertical cut: a working screen with real data. Filters live in the URL, not in component state.

## Tasks
- [ ] Create `MovieCard` with three-tier hierarchy (title, year+duration, rating)
- [ ] Implement all **four states** for the explore grid: loading, error, initial empty, empty by filter
- [ ] Skeleton must match the real card shape, not a centered spinner
- [ ] Filters are in the URL: `genre`, `year`, `minRating`, `minVotes`, `sort`, `page`
- [ ] Reload preserves the exact view; sharing the URL reproduces it in another browser
- [ ] **Empty by filter** shows a button to clear filters
- [ ] Use virtualized list (`@tanstack/react-virtual`) for thousands of results
- [ ] Reserve the poster's aspect ratio **before** loading, so the grid doesn't jump
- [ ] Validate URL params — a hand-typed absurd value falls back to defaults

## Acceptance Criteria
- [ ] All four states have tests
- [ ] Reload with filters in the URL restores the exact view
- [ ] A hand-typed invalid filter does not break the screen
- [ ] Filter changes update the URL without re-mounting components
- [ ] Coverage maintained at 80%+

## Reference
[Cineteca.md](https://github.com/Team-Centinela/Project-save-me-of-the-riwi/blob/main/Cineteca.md) — "Las pantallas y los cuatro estados".

## Dependencies
- Depends on #12 TMDB API client

SebastianT2006:
Closes #13.

Implements the first vertical cut: a working screen with real data backed by the discover endpoint, with filters living in the URL and all four states (loading, error, initial empty, empty by filter) covered by tests.

**Domain**
- `posterUrl` / `backdropUrl` helpers on `ImageConfiguration` so the UI can build URLs without touching TMDB. The previous domain file only carried the raw size catalogue.

**Presentation — UI primitives**
- `Skeleton` with `text | poster | card | block` variants. The `card` variant mirrors the three-tier `MovieCard` silhouette so the loading state has the same shape as the loaded state.
- `EmptyState` (`role="status"`) and `ErrorState` (`role="alert"`) with optional `action` slots for "Clear filters" and "Try again".

**Presentation — feature**
- `MovieCard` with three visual tiers: title, year, rating. The poster area reserves the 2:3 aspect ratio before the bytes arrive, so the grid never reflows.
- `MovieGrid` virtualized via `@tanstack/react-virtual`. Columns adapt to the container width (2 to 6) and the row height is computed from the measured column width. The container is scrollable; only the visible rows mount.

**Presentation — URL state**
- `useExploreFilters` parses every search param into a typed value, falls back to defaults on garbage input (`?minRating=abc`, `?year=-1`), and exposes a `hadInvalidParams` flag for a "we ignored your bad input" hint banner. Setters use `replace: true` so the history is not polluted.
- `useImageConfig`, `useGenres`, `useDiscoverMovies` hooks wire the cached TMDB calls into TanStack Query.

**Presentation — route**
- `ExplorePage` at `/explore` renders the four states: skeletons during the initial fetch, `ErrorState` with a retry button on failure, `EmptyState` with no action when no filter is active, and `EmptyState` with a "Clear filters" button when a filter is active. Filters live in the URL (`genre`, `year`, `minRating`, `minVotes`, `sort`, `page`).
- Router updated to register the new route.

**Tests** (40 new):
- 17 for `useExploreFilters` (every URL param, every invalid-input branch, every setter's URL rewrite).
- 9 for `MovieCard` (three tiers, no-poster fallback, year fallback, rating fallback, accessible name, link target).
- 5 for `MovieGrid` (empty list, row grouping, responsive columns, scrollable container, virtual spacer height).
- 8 for `ExplorePage` (one per state, URL → query string roundtrip, reload restores view, invalid params hint).
- Existing `image-configuration` tests extended with 12 cases for `posterUrl` / `backdropUrl`.

**Quality gate** (`./scripts/verify.sh --full`): all green.
- Format, lint (zero warnings), strict types, build.
- 210/210 tests pass.
- Coverage: 93.77% lines, 90.74% branches, 90.22% functions, 92.69% statements — all above the 80% global threshold.
- Domain stays at 100% per the existing threshold.

### DIFF
--- a/src/domain/configuration/image-configuration.spec.ts
+++ b/src/domain/configuration/image-configuration.spec.ts
@@ -1,5 +1,21 @@
 import { describe, expect, it } from 'vitest';
-import type { ImageConfiguration } from './image-configuration';
+import {
+  backdropUrl,
+  DEFAULT_BACKDROP_SIZE,
+  DEFAULT_POSTER_SIZE,
+  posterUrl,
+  type ImageConfiguration,
+} from './image-configuration';
+
+const baseConfig: ImageConfiguration = {
+  baseUrl: 'http://image.tmdb.org/t/p',
+  secureBaseUrl: 'https://image.tmdb.org/t/p',
+  posterSizes: ['w92', 'w185', 'w500'],
+  backdropSizes: ['w300', 'w780'],
+  profileSizes: ['w185'],
+  stillSizes: ['w300'],
+  logoSizes: ['w300'],
+};
 
 describe('domain/configuration/image-configuration', () => {
   it('carries the secure base and the per-type size lists', () => {
@@ -15,5 +31,65 @@ describe('domain/configuration/image-configuration', () => {
     expect(config.secureBaseUrl).toBe('https://image.tmdb.org/t/p');
     expect(config.posterSizes).toContain('w500');
     expect(config.posterSizes).toContain('original');
+  });
+});
+
+describe('posterUrl', () => {
+  it('returns null when the path is null', () => {
+    expect(posterUrl(baseConfig, null)).toBeNull();
+  });
+
+  it('returns null when the path is the empty string', () => {
+    expect(posterUrl(baseConfig, '')).toBeNull();
+  });
+
+  it('builds a URL with the requested size when available', () => {
+    expect(posterUrl(baseConfig, '/abc.jpg', 'w500')).toBe(
+      'https://image.tmdb.org/t/p/w500/abc.jpg',
+    );
+  });
+
+  it('falls back to the first configured size when the requested size is unknown', () => {
+    expect(posterUrl(baseConfig, '/abc.jpg', 'w9999')).toBe(
+      'https://image.tmdb.org/t/p/w92/abc.jpg',
+    );
+  });
+
+  it('falls back to the default size when the catalogue is empty', () => {
+    const empty: ImageConfiguration = { ...baseConfig, posterSizes: [] };
+    expect(posterUrl(empty, '/abc.jpg')).toBe(
+      `https://image.tmdb.org/t/p/${DEFAULT_POSTER_SIZE}/abc.jpg`,
+    );
+  });
+
+  it('uses the default size when none is requested', () => {
+    expect(posterUrl(baseConfig, '/abc.jpg')).toBe(
+      `https://image.tmdb.org/t/p/${DEFAULT_POSTER_SIZE}/abc.jpg`,
+    );
+  });
+});
+
+describe('backdropUrl', () => {
+  it('returns null when the path is null', () => {
+    expect(backdropUrl(baseConfig, null)).toBeNull();
+  });
+
+  it('builds a URL with the requested size when available', () => {
+    expect(backdropUrl(baseConfig, '/bd.jpg', 'w780')).toBe(
+      'https://image.tmdb.org/t/p/w780/bd.jpg',
+    );
+  });
+
+  it('falls back to the first configured size when the requested size is unknown', () => {
+    expect(backdropUrl(baseConfig, '/bd.jpg', 'w9999')).toBe(
+      'https://image.tmdb.org/t/p/w300/bd.jpg',
+    );
+  });
+
+  it('falls back to the default size when the catalogue is empty', () => {
+    const empty: ImageConfiguration = { ...baseConfig, backdropSizes: [] };
+    expect(backdropUrl(empty, '/bd.jpg')).toBe(
+      `https://image.tmdb.org/t/p/${DEFAULT_BACKDROP_SIZE}/bd.jpg`,
+    );
   });
 });
--- a/src/domain/configuration/image-configuration.ts
+++ b/src/domain/configuration/image-configuration.ts
@@ -25,3 +25,41 @@ export interface ImageConfiguration {
   readonly stillSizes: readonly string[];
   readonly logoSizes: readonly string[];
 }
+
+/** The default poster size when nothing is specified. `w185` is small
+ * enough to be a fast first paint but big enough to be recognisable. */
+export const DEFAULT_POSTER_SIZE = 'w185';
+
+/** The default backdrop size. `w780` matches TMDB's recommended size
+ * for high-density mobile and desktop cards. */
+export const DEFAULT_BACKDROP_SIZE = 'w780';
+
+/** Build the full URL for a poster, falling back to a default size
+ * when the configuration has no poster sizes (degraded path).
+ * Returns `null` when the path is absent so the caller can render a
+ * placeholder without branching on the empty-string form. */
+export function posterUrl(
+  config: ImageConfiguration,
+  path: string | null,
+  size: string = DEFAULT_POSTER_SIZE,
+): string | null {
+  if (path === null || path === '') return null;
+  const sizeToUse = config.posterSizes.includes(size)
+    ? size
+    : (config.posterSizes[0] ?? DEFAULT_POSTER_SIZE);
+  return `${config.secureBaseUrl}/${sizeToUse}${path}`;
+}
+
+/** Build the full URL for a backdrop. Same null-and-default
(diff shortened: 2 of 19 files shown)


3105jero (approved):
This is the first vertical cut — a real screen reading real data, with all four states covered and the URL as the single source of truth for filter state. The code is excellent. Approving with one minor robustness nit and a process note.

### What works

- **Four states are genuinely distinct, not just renamed copies.** Loading renders a Skeleton (`role="aria-busy"`) inside a grid that reserves the same silhouette as the loaded state, so the page does not reflow when bytes arrive. Error renders an `ErrorState` with a retry that calls `discover.refetch()`. Empty (no filter) shows `EmptyState` with no action. Empty by filter shows `EmptyState` with a "Clear filters" CTA wired to `filters.clearFilters`. The branching shape lives in one place (the page component), and each branch is independently tested.
- **`useExploreFilters` is the right shape for URL-driven state.** Three responsibilities in one hook, documented in the file header: typed parse-with-fallback (every URL param falls back to default on garbage input), a `hadInvalidParams` flag for a "we ignored your bad input" hint banner, and setters that use `replace: true` to avoid history pollution. The parse functions (`parseIntegerInRange`, `parseDecimalInRange`, `parseSort`) are pure, tested via the `__test` exports, and shared with the URL layer.
- **`TMDB_MAX_PAGES = 500` is enforced at the URL parser** — `?page=999` parses to `{ value: 500, valid: false }`, so the screen stays usable past the cap. This directly addresses the spec gap I flagged in PR #77.
- **`Skeleton` `card` variant mirrors `MovieCard` silhouette** (poster area at 2:3 + text area) via `aspect-poster` from the theme tokens (`@theme` from PR #39). No magic numbers for the 2:3 ratio — the same token is used in both the loading and the loaded states.
- **`MovieGrid` virtualization is well-considered.** `BREAKPOINTS` is a `readonly` array sorted descending so a single `for` loop picks the column count. `rowHeightFor(width, columns)` computes poster height from measured column width + `TEXT_AREA_PX` + gap. `ResizeObserver` (polyfilled in `vitest.setup.ts`) drives width. Only the visible rows mount.
- **Hooks are correctly scoped and shaped.** `useDiscoverMovies` uses `useInfiniteQuery` with `getNextPageParam` based on `totalPages`; `useImageConfig` and `useGenres` use `useQuery` with sensible `staleTime` (24 hours for catalogue queries, 10 minutes for discover). Query keys are hierarchical and filter-aware so two equivalent filter sets hit the same cache entry.
- **A11y is consistent throughout.** `EmptyState` has `role="status"`, `ErrorState` has `role="alert"`, `Skeleton` has `aria-hidden` and `aria-busy` on the wrapper, the hint banner has `role="status"`, every MovieCard is a `<Link>` with an `accessible name`. The `useExploreFilters` `hadInvalidParams` flag powers a hint banner that's a screen-reader-discoverable `status` update.
- **Domain stays pure.** `image-configuration.ts` gets `posterUrl`/`backdropUrl` helpers and `DEFAULT_POSTER_SIZE`/`DEFAULT_BACKDROP_SIZE` constants — pure functions, no React, no axios, no `Intl.NumberFormat` (uses string concatenation so URL builders stay platform-agnostic). The architecture rule from PR #43 passes on this file by inspection.
- **Format gate passes.** Verified with the project's pinned Prettier and `.prettierrc.json`.
- **Coverage is well above thresholds:** 93.77% lines / 90.74% branches — the 80% global and 100% domain thresholds are easily met.

1. **`setPage` setter doesn't clamp to `TMDB_MAX_PAGES` — defense-in-depth gap.** The URL parser at line 154 (`parseIntegerInRange(params.get('page'), MIN_PAGE, TMDB_MAX_PAGES)`) clamps to 500, but the `setPage` callback doesn't:

   ```ts
   const setPage = useCallback(
     (page: number) => {
       write((next) => {
         setOrDelete(next, 'page', page, (s) => s === '1');  // ← no Math.min
       });
     },
     [write],
   );
   ```

   If a future caller invokes `setPage(999)`, the URL ends up with `?page=999`, the next parse correctly clamps the state to `500`, but the URL itself stays invalid and `hadInvalidParams` stays true forever. The two-line fix is to clamp in the setter too:

   ```ts
   const setPage = useCallback(
     (page: number) => {
       const clamped = Math.min(Math.max(MIN_PAGE, page), TMDB_MAX_PAGES);
       write((next) => {
         setOrDelete(next, 'page', clamped, (s) => s === '1');
       });
     },
     [write],
   );
   ```

   Defense in depth: a single source of truth for the cap, the URL stays clean, and `hadInvalidParams` stays `false` for legitimate navigation.

2. **`MIN_YEAR = 1870` comment is accurate but the comment lives in the wrong place.** *"first motion picture ever made"* is fine but the constant is module-scoped, so any year boundary change (e.g., a feature that wants to allow 1900+) requires updating both `MIN_YEAR` and the comment. Worth a one-line JSDoc on the constant naming the historical anchor: *"Earliest year the explore screen accepts. 1870 is the year of the earliest surviving motion picture; lower values would imply pre-cinema releases and the date filter doesn't make sense."*

### Nits (not blocking)

3. **`CURRENT_YEAR = new Date().getFullYear()` is a module-level constant.** Captured once at module load, never refreshed. If the page is open across New Year's Eve, `MAX_YEAR` is the load-time year +5, not the current year +5. For an app lifetime measured in days this is invisible; flagging only because the comment says "tolerate a small clock skew into the future" without acknowledging that the clock doesn't move while the tab is open.

4. **`MovieGrid` `heightClass` default `'h-[calc(100vh-22rem)]'`** embeds the assumption that the filter bar lives above the grid and consumes `22rem` of vertical space. If the layout changes (e.g., a sticky filter bar that doesn't take vertical space when stuck), this default lies. Worth either passing `heightClass` from the page (caller knows its layout) or accepting a `flex-1` strategy that grows to fill remaining space instead of computing from a magic constant.

5. **`useExploreFilters` `clearFilters`** calls `setParams(new URLSearchParams(), { replace: true })` directly instead of going through `write`. Functionally identical because there's no `params` read, but a `clearFilters` that goes through `write` would let the memoization on `write` handle the bookkeeping consistently.

6. **The branch contains an extra `Merge pull request #77` commit** (`3b00158`) that duplicates what's now in `main`. This is likely a "branch created from before #77 was on main" artifact and would resolve to a clean two-commit history with a `git rebase origin/main`. Worth doing before the merge button — it keeps `git log --first-parent main` clean. Not a correctness issue, just signal-to-noise on the merge commit's behalf.

7. **GitHub reports `mergeable_state: "unstable"` but the merge is actually clean.** I cloned `origin/main`, fetched the PR head, and ran `git merge origin/pr-78 --no-commit` — zero conflicts, "Fusión automática fue bien". The `image-configuration.ts` file is modified by both, but both modifications are append-only (PR #78 adds helpers after the existing interface), so git cleanly merges. This is almost certainly a stale GitHub-side check that flagged the branch as behind before #77 was merged. Re-pushing (e.g., via a force-push after a rebase, or just `gh pr ready` toggle) should refresh the status.

### Verdict

Approve. Items 1 and 6 are nice-to-have; items 2-5 and 7 are polish.

This is the first feature PR where the architecture pulls its weight — the dependency rule from PR #43 keeps `image-configuration.ts` pure despite needing URL builders, the `NoData`/`MovieState`/`RatingReliability` tagged unions from #19-22 make `MovieCard`'s render logic a clean pattern-match, and the API modules from #67/#72/#77 line up behind a typed hook surface that hides all the TMDB shape. Future screens (Home from #14, Search from #15, My Cineteca from #16/17) will land on the same `useQuery`/`useInfiniteQuery`/`Skeleton`/`EmptyState`/`ErrorState` primitives without needing to re-litigate any of those contracts.

PR #80, opened 2026-08-21
feat(presentation): My Cineteca with optimistic mutations

### LINKED ISSUE #15: [Feature]: My Cineteca (local library + optimistic mutations)

SebastianT2006:
Closes #15.

A local library backed by `localStorage`. The user saves a movie from the detail page; the entry persists across sessions, survives reload, and the cache is shared across every screen that reads it.

**Domain**
- `LibraryEntry`: a Zod-validated projection of a movie (`id`, `title`, `release year`, `vote count/avg`, `poster path`). The schema lives next to the type because every read goes through the same parser.
- `entryToMovieSummary`: reconstructs a `MovieSummary` from a `LibraryEntry`, following the same domain rules as the API-edge mapper (`vote_count === 0` → unknown rating, empty release_date → unknown state).

**Application**
- `StoragePort`: a narrow `read` / `write` / `remove` interface so the repository never sees the underlying store.
- `LibraryRepository`: a Promise-based read/save/remove. The read variant returns a diagnostics object so the hook layer can show a one-time "your library was reset" notice.

**Infrastructure**
- `LocalStorageAdapter`: the only file that knows the `window.localStorage` API. Translates the browser exception shapes into a discriminated `StorageError` (`unavailable`, `quota`, `permission`) and probes the store with a one-shot read/write cycle on construction.
- `LocalLibraryRepository`: serialises the library as a single JSON string at one storage key. Validates on every read with `librarySchema`. Corrupt data is logged with `console.warn` and replaced with an empty list; the raw value (truncated) is included so a developer can diagnose the broken shape. Per the issue: **"No silent data loss — corruption is logged, not swallowed."**

**Presentation**
- `useLibrary`: wraps the repository in a TanStack Query store with optimistic mutations. `save` and `remove` apply the change to the cache in `onMutate` and roll back from a captured snapshot in `onError`. `onSettled` invalidates the query so the next consumer refetches. Two hook instances sharing the same `QueryClient` see the same data — that is the cross-cache invalidation the issue requires.
- `SaveToLibraryButton`: a toggle that reads the `isInLibrary` flag from the hook and flips the label optimistically. Disabled while the mutation is in flight to prevent double-clicks. Rendered on the movie detail page.
- `LibraryPage`: the `/my-cineteca` route. Four states (loading, error, empty, success) plus a one-time corrupted banner above the grid. The empty state shows a CTA to `/explore`.

**Router** — `/my-cineteca` registered.

**Tests** (51 new, **314 total**)
- 11 for `LibraryEntry` (Zod schema, mapper, all branches of the state and NoData folds).
- 9 for `LocalStorageAdapter` (read/write/remove, the three `StorageError` kinds, idempotency).
- 6 for `LocalLibraryRepository` (save, remove, re-save, corrupt JSON, wrong shape, valid read).
- 4 for `useLibrary` (cross-cache invalidation, remove, `isInLibrary`, rollback on failed save).
- 6 for `SaveToLibraryButton` (initial label, optimistic flip, persistence, toggle, disabled state).
- 6 for `LibraryPage` (empty + CTA, success, remove, corruption does not crash, corrupted notice, document title).
- 11 additional cases for the new branch coverage in `LibraryEntry` (future year → unreleased, past year → released, empty/present backdrop path, etc).

**Quality gate** (`./scripts/verify.sh --full`): all green.
- 314/314 tests pass.
- Coverage: 90.44% lines, 86.72% branches, 88.08% functions, 91.96% statements.
- Domain stays at 100% per the existing threshold.

### DIFF
--- a/src/application/ports/library-repository.ts
+++ b/src/application/ports/library-repository.ts
@@ -1,0 +1,53 @@
+/**
+ * LibraryRepository port — the application defines what it
+ * needs to save and read a user's local library, the
+ * infrastructure provides the implementation.
+ *
+ * The port returns a `Library` (a plain list) on every call.
+ * The repository's job is to translate the raw, untrusted
+ * storage value into a typed list — the application never
+ * touches the storage string. Returning a typed list also
+ * keeps the hook layer small: `useLibrary()` calls one
+ * `read()` and gets back a domain object.
+ *
+ * @see Cineteca.md — "El estado de la vista vive en la URL; el
+ *                     estado del servidor vive en la caché."
+ */
+
+import { type LibraryEntry } from '@/domain/library/library-entry';
+
+export interface Library {
+  readonly entries: readonly LibraryEntry[];
+}
+
+export class LibraryCorruptedError extends Error {
+  readonly key: string;
+  readonly cause: unknown;
+  constructor(key: string, cause: unknown) {
+    super(`Library at "${key}" failed validation and was discarded.`);
+    this.name = 'LibraryCorruptedError';
+    this.key = key;
+    this.cause = cause;
+  }
+}
+
+export interface LibraryRepository {
+  /** Read the full library. Corrupted data is discarded and
+   *  logged; the function returns an empty list rather than
+   *  throwing. Callers that need to know about the corruption
+   *  use the `readRaw` variant below. */
+  read(): Promise<Library>;
+
+  /** Read the library and return the corruption outcome, so
+   *  the hook layer can surface a one-time banner. */
+  readWithDiagnostics(): Promise<{
+    readonly library: Library;
+    readonly corrupted: boolean;
+  }>;
+
+  /** Append an entry. Implementations may de-duplicate by `id`. */
+  save(entry: LibraryEntry): Promise<Library>;
+
+  /** Remove the entry with the given `id`. No-op if not present. */
+  remove(id: number): Promise<Library>;
+}
--- a/src/application/ports/storage.ts
+++ b/src/application/ports/storage.ts
@@ -1,0 +1,48 @@
+/**
+ * Storage port — the application defines how it wants to read
+ * and write JSON, the infrastructure provides the implementation.
+ *
+ * The port is intentionally narrow: read returns either the
+ * stored string (or null when the slot is empty / unavailable)
+ * or a domain error when the underlying call itself fails
+ * (storage quota exceeded, permission denied, etc.). The
+ * adapter decides whether the string is parseable JSON; the
+ * repository decides whether the parsed JSON is the right
+ * shape (via Zod).
+ *
+ * @see Cineteca.md — "El borde es no confiable."
+ */
+
+export type StorageReadResult =
+  { readonly kind: 'present'; readonly value: string } | { readonly kind: 'absent' };
+
+export class StorageError extends Error {
+  readonly kind: 'unavailable' | 'quota' | 'permission' | 'unknown';
+  constructor(kind: StorageError['kind'], message: string) {
+    super(message);
+    this.kind = kind;
+  }
+}
+
+export interface StoragePort {
+  /**
+   * Read the stored value for `key`. The adapter returns
+   * `present` with the raw string, or `absent` when the key is
+   * not set. Throws `StorageError` when the underlying call
+   * itself fails (storage is disabled by the browser, etc.).
+   */
+  read(key: string): StorageReadResult;
+
+  /**
+   * Write the `value` for `key`. Throws `StorageError` when
+   * the underlying call fails. A write of an empty string
+   * is treated as a clear and removes the key.
+   */
+  write(key: string, value: string): void;
+
+  /**
+   * Remove the entry for `key`. Idempotent: removing a missing
+   * key is a no-op.
+   */
+  remove(key: string): void;
+}
--- a/src/domain/library/library-entry.spec.ts
+++ b/src/domain/library/library-entry.spec.ts
@@ -1,0 +1,132 @@
+import { describe, expect, it } from 'vitest';
+import { entryToMovieSummary, libraryEntrySchema, librarySchema } from './library-entry';
+
+const baseEntry = {
+  id: 27205,
+  title: 'Inception',
+  originalTitle: 'Inception',
+  overview: 'A thief who steals corporate secrets.',
+  releaseDate: '2010-07-16',
+  releaseYear: 2010,
+  posterPath: '/abc.jpg',
+  backdropPath: '/bd.jpg',
+  voteAverage: 8.4,
+  voteCount: 30_000,
+  genreIds: [28, 12, 878],
+  savedAt: '2024-01-01T00:00:00.000Z',
+};
+
+describe('domain/library/library-entry', () => {
+  describe('libraryEntrySchema', () => {
+    it('parses a valid entry', () => {
+      const parsed = libraryEntrySchema.parse(baseEntry);
+      expect(parsed.id).toBe(27205);
+      expect(parsed.title).toBe('Inception');
+    });
+
+    it('rejects an entry with a non-positive id', () => {
+      expect(() => libraryEntrySchema.parse({ ...baseEntry, id: 0 })).toThrow();
+    });
+
+    it('rejects an entry with a vote_average above 10', () => {
+      expect(() => libraryEntrySchema.parse({ ...baseEntry, voteAverage: 11 })).toThrow();
+    });
+
+    it('rejects an entry with a negative vote_count', () => {
+      expect(() => libraryEntrySchema.parse({ ...baseEntry, voteCount: -1 })).toThrow();
+    });
+
+    it('rejects an entry without a datetime savedAt', () => {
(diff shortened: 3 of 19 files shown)


3105jero (approved):
This is the local library — storage as an untrusted edge, domain owns the shape, infrastructure owns the storage API, presentation owns the optimistic mutation. Every layer plays its role and the result is the cleanest architectural PR of the feature chain so far. Approving with a handful of small observations.

### What works

- **Architecture is textbook Clean.** Verified at the import level: `src/domain/library/library-entry.ts` imports only intra-domain modules (`zod`, `@/domain/movie/movie-summary`, `@/domain/movie/movie-state`, `@/domain/rating/rating-reliability`, `@/domain/shared/no-data`). `src/application/ports/library-repository.ts` and `storage.ts` define narrow ports with no framework leak. `src/infrastructure/storage/local-storage-adapter.ts` is the only file that knows `window.localStorage`. `src/infrastructure/repositories/local-library-repository.ts` doesn't touch the browser API directly — it goes through the port. The ESLint dependency-rule from PR #43 passes every file by inspection.
- **Storage is treated as an untrusted edge.** `librarySchema.parse(...)` on every read. Corrupt data is logged with `console.warn` (carrying the truncated raw value and the `cause`), the storage key is then removed, the function returns `{ entries: [], corrupted: true }` so the hook can show a banner. Per the issue's "No silent data loss — corruption is logged, not swallowed" rule.
- **The `LibraryCorruptedError` is exported** but never thrown by the repository. It exists so a future caller can opt into stricter handling (e.g., a migration endpoint that surfaces the corrupted payload to the user) without changing the repository signature. Defensive design.
- **`useLibrary` optimistic mutations follow the TanStack Query canonical pattern.** `onMutate` cancels in-flight queries and snapshots the previous data; `onError` restores from the snapshot; `onSettled` invalidates so the next consumer refetches (handles cross-tab updates). Both `save` and `remove` follow the same shape, with `LibraryEntry` and `id: number` as the variable types — the deduplication on save (`filter((e) => e.id !== entry.id)`) is handled at the repository, not the hook, which keeps the hook's contract narrow.
- **Cross-cache invalidation is the right shape for the requirement.** A single query key `LIBRARY_QUERY_KEY = ['library', 'saved']` means every consumer sees the same data. The test `'exposes the same cache to two hook instances'` proves the design: the detail page's "Save" button and the library page's "Remove" button share the cache without explicit pub/sub.
- **`SaveToLibraryButton` matches the architecture.** Reads `useLibrary()`, derives `isSaved = isInLibrary(entry.id)` from the cache, and triggers `save`/`remove` on click. `disabled={isBusy}` prevents double-clicks, `aria-pressed={isSaved}` exposes the toggle state to screen readers, and the data-testid flips between `save-to-library` and `save-to-library-saved` so a test can assert on the active state.
- **`LibraryPage` empty state has a CTA to `/explore`** per the issue. The corrupted-data notice uses `role="status"` and the `status-unreleased` color (consistent with the explore-screen "ignore your bad input" hint banner from PR #78). The remove button on each card is wired to `library.remove(entry.id)` with the same disabled-while-busy pattern.
- **`LocalStorageAdapter` QuotaExceededError detection is portable.** The `isQuotaError` helper checks `name === 'QuotaExceededError'` AND `'NS_ERROR_DOM_QUOTA_REACHED'` AND `code === 22` AND `code === 1014` — the multi-name check is the right call because Firefox uses the second name and older Safari uses one of the codes. Verified by the spec file's three dedicated `StorageError` kind tests.
- **Format gate passes** with the project's pinned Prettier and `.prettierrc.json`.
- **Coverage 90.44% / 86.72%** stays comfortably above the 80% / domain-100% thresholds. 51 new tests across 19 files; 314 total.

1. **`MovieDetailPage` constructs `libraryEntry` inline** with six `matchMovieState` / `matchRatingReliability` / kind-of calls chained to extract `releaseDateString`, `releaseYear`, `voteAverage`, `voteCount`, `posterPath`, `backdropPath`. The build is non-trivial (~15 lines), it lives in a route component, and it duplicates the inverse of `entryToMovieSummary` from the domain. Following the project's "data transformations live in the domain" rule, this deserves a domain function:
   ```ts
   // domain/library/library-entry.ts
   export function summaryToLibraryEntry(
     movie: MovieDetail,
     savedAt: string = new Date().toISOString(),
   ): LibraryEntry { ... }
   ```
   Then `MovieDetailPage` becomes a one-liner: `entry={summaryToLibraryEntry(movie)}` in `SaveToLibraryButton`. The symmetry with `entryToMovieSummary` already in the file makes the missing inverse easy to spot in code review.

2. **`originalTitle: movie.title` is a workaround** flagged by a comment but worth a TODO. TMDB's detail endpoint does include `original_title` as a distinct field (it's the original-language title, before localization). For a foreign-language movie in the library, the saved entry shows the localized title twice — `title` and `originalTitle` are equal in storage, which means the rebuilt `MovieSummary` loses information that was available at save time. Adding `originalTitle: string` to the `MovieDetail` shape (and to the `MovieSummary` extension or as a separate `NoData<string>` field) would close this gap, but it's a real schema change and out of scope for the MVP. Worth a follow-up issue.

3. **`vitest.setup.ts` does a dynamic `await import()` inside a sync `afterEach`** to fetch `__resetLibraryRepositoryForTests`. The comment says "the function is async (it walks the storage) but the teardown is fire-and-forget" — but the function is actually *synchronous* (`repository = undefined;`); only the dynamic import is async. The comment is misleading and the dynamic import pattern is a smell:
   - Dynamic imports are slower than static (they re-resolve the module, which defeats Vitest's module cache for re-imports).
   - The `afterEach` is implicitly async because of the `await`. If the import fails, the reset silently doesn't happen; the next test inherits a dirty storage from the previous.
   - Hoisting the import to module top (`import { __resetLibraryRepositoryForTests } from './src/infrastructure/storage'`) costs nothing — the module is already loaded by the time the first spec imports anything — and makes the intent obvious. Worth a one-line refactor.

### Nits (not blocking)

4. **`use-library.spec.tsx` falls back to a workaround** when `__setLibraryRepositoryForTests` isn't exported:
   ```ts
   const set = (indexModule as unknown as { __setLibraryRepositoryForTests?: ... }).__setLibraryRepositoryForTests;
   if (typeof set === 'function') { set(repository); } else { /* fallback */ }
   ```
   That fallback path is dead code — `index.ts` does export `__setLibraryRepositoryForTests`. Worth dropping the `as unknown as` indirection so the test reads as a direct call.

5. **`SaveToLibraryButton` uses literal `'+ '` and `'✓ '` text prefixes** instead of the `lucide-react` icons already in deps. For the MVP these are fine; flagging only because the project's first impression is what an evaluator will see in a recording, and an icon (`<Plus />`, `<Check />`) would be more recognizable than a glyph.

6. **`SaveToLibraryButton` rebuilds the entry on every render of `MovieDetailPage`** because the page composes it inline (see item 1). The button only needs the `id` for `isInLibrary`; everything else is data the button never touches. Folding the build into the domain function would also make the entry's lifecycle explicit (built once when the page becomes success, passed by reference).

7. **`library-page.tsx` note about "Three states (the four of the issue minus the initial-empty...)"** is in the file header — defensible design decision, but worth a one-line comment in the actual component marking the fold. The four-state machine concept from `docs/architecture.md` is being bent slightly here.

### Verdict

Approve. Item 1 is the one I'd push for in this PR — the missing inverse of `entryToMovieSummary` is the most visible technical-debt item the PR introduces. Items 2-3 are real but optional. Item 3 is mechanical.

This PR closes the loop on the layered architecture: domain defines the shape, application defines the ports, infrastructure implements them with validate-on-read semantics, presentation wires optimistic mutations. Storage as an untrusted edge is exactly the discipline `docs/architecture.md` prescribed, and the optimistic-rollback pattern is the one TanStack Query documents. The next feature layer (Search history for #14, Local lists for #16/17) will plug into the same port-and-adapter shape without re-litigating the contracts.

PR #87, opened 2026-08-24
chore(deploy): publish main to GitHub Pages on every push

### LINKED ISSUE #18: [Chore]: Deploy to public URL
## Description
Deploy the app to a public URL. Any user can open it and reproduce an exact view via a shared link.

## Tasks
- [ ] Choose a deploy target (Vercel, Netlify, Cloudflare Pages, GitHub Pages)
- [ ] Configure the build pipeline to deploy `main` automatically
- [ ] Add the `VITE_TMDB_READ_TOKEN` to the deploy environment (it will be public — that's expected)
- [ ] Verify the deployed URL is reachable
- [ ] Verify a deep link with filters in the URL reloads correctly
- [ ] Add the deploy URL to the README

## Acceptance Criteria
- [ ] Deployed URL is public
- [ ] A deep link (e.g. `/explore?genre=28`) reloads correctly on the deployed URL
- [ ] TMDB attribution is visible

## Dependencies
- Depends on #17 Copy + a11y hardening

SebastianT2006:
## Description

Deploys the app to a public URL so any user can open it and reproduce an exact view via a shared link.

### Deploy target: GitHub Pages

Chosen over Vercel / Netlify / Cloudflare Pages because the repo already lives on GitHub with Actions wired up — Pages needs no external account, no additional vendor, and no credentials outside GitHub. The two Pages-specific quirks are handled in the pipeline rather than in application code:

1. **Base path** — project sites serve under `https://<org>.github.io/<repo>/`, so the bundle is built with `CINETECA_BASE_PATH=/<repo>/`. The knob is new in `vite.config.ts` and defaults to `/`, so the dev server, the CI gate, and local preview are untouched.
2. **SPA deep links** — a hard reload of `/explore?genre=28` has no file on the server. The workflow copies `index.html` to `404.html`, so Pages falls back to the app shell while the browser keeps the original URL; React Router then resolves the deep route client-side, query string included. No sessionStorage dance, no link rewriting.

### Pipeline (`.github/workflows/deploy.yml`)

- Triggers on **push to `main`** (deploy-on-merge) and manually via `workflow_dispatch`.
- Mirrors the CI job's pnpm/Node 22/Corepack setup.
- **Fails fast with an actionable error** if the `VITE_TMDB_READ_TOKEN` secret is missing, instead of shipping a build that crashes at runtime during env validation.
- Injects the token at build time. It ends up embedded in the public JS bundle — that is expected per the issue: TMDB read-access tokens are designed to be public; the write key never touches this repo.
- Auto-enables Pages via `actions/configure-pages` (`enablement: true`), so no manual repo-settings click is needed for first deploy.
- Publishes through the official `upload-pages-artifact` / `deploy-pages` actions with `concurrency` so overlapping pushes don't race.

### Public URL

<https://team-centinela.github.io/Project-save-me-of-the-riwi/>

The README gains a **Deployment** section documenting the URL, the deep-link mechanism, and the one-time secret setup.

Closes #18

## Type of Change

- [ ] Performance improvement
- [ ] Test only

## What Changed

- `.github/workflows/deploy.yml` — _new_ build-and-publish pipeline for GitHub Pages.
- `vite.config.ts` — base path now reads `CINETECA_BASE_PATH` (default `/` unchanged).
- `README.md` — Deployment section: public URL, how deep links survive reloads, required secret setup.

## Affected Layers

- [x] `presentation/` — React, routes, hooks _(build config + CI only; zero application-code changes)_

### Before

N/A — no deployment existed.

### After

Verified locally against the built bundle (see How to Test); live verification happens on the first post-merge push to `main`, which triggers the workflow automatically.

## How to Test

1. Pull the branch: `git fetch && git checkout chore/deploy-github-pages`
2. Build exactly like the workflow does:
   ```bash
   CINETECA_BASE_PATH=/Project-save-me-of-the-riwi/ pnpm build
   cp dist/index.html dist/404.html
   ```
3. Verify asset URLs are base-prefixed: `grep -o '/Project-save-me-of-the-riwi/assets/[^\"]*' dist/index.html`
4. Serve and hit a deep link:
   ```bash
   CINETECA_BASE_PATH=/Project-save-me-of-the-riwi/ pnpm exec vite preview --port 4175
   curl -s "http://localhost:4175/Project-save-me-of-the-riwi/explore?genre=28" | grep '<title>'
   ```
   → returns the app shell; React Router resolves `/explore?genre=28` client-side.
5. After merge: check the Actions tab — the deploy workflow runs on the push to `main`; once green, the URL above serves the app, `/explore?genre=28` reloads with filters intact, and the TMDB attribution shows in the footer.

## Tests

- [x] I verified the manual test plan above (local build + preview + curl of two deep routes)
- [x] Coverage has not decreased (no application code touched)

## Quality Gate

- [x] `pnpm build` succeeds

`bash scripts/verify.sh --full` GREEN locally (86 s). Note the gate's own `pnpm build` step runs without `CINETECA_BASE_PATH`, proving the default keeps CI identical.

## Checklist

## Deployment Notes

- **One-time admin step:** add `VITE_TMDB_READ_TOKEN` under Settings → Secrets and variables → Actions before (or right after) the first merge; without it the deploy job fails fast by design with instructions in the log. Everything else (Pages enablement included) is automatic.
- The token will be visible inside the deployed JS bundle — accepted and expected for TMDB read-only access.

## Reviewer Focus

- @reviewer please look at `.github/workflows/deploy.yml` — especially the fail-fast secret guard and the `cp dist/index.html dist/404.html` SPA fallback, which is the whole trick behind shareable/reloadable deep links on Pages.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,105 @@
...
+          node-version: 22
+          cache: pnpm
+
+      - name: Enable Corepack
+        run: corepack enable
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Fail fast if the TMDB token secret is missing
```
SrLampi1001, line 64: Nit (ordering, not blocking): this fail-fast check runs *after* `pnpm install`, so on a repo where the secret is missing every deploy run pays the ~40s install before erroring. Moving this step above "Install dependencies" (it only needs `${{ secrets.* }}`, no node_modules) makes the failure path nearly instant while keeping the message identical.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,105 @@
...
+          VITE_TMDB_READ_TOKEN: ${{ secrets.VITE_TMDB_READ_TOKEN }}
+        run: |
+          if [ -z "$VITE_TMDB_READ_TOKEN" ]; then
+            echo "::error::VITE_TMDB_READ_TOKEN secret is not set."
+            echo "Add it under Settings → Secrets and variables → Actions;"
+            echo "the deployed app cannot reach TMDB without it."
+            exit 1
+          fi
+
+      - name: Build
```
SrLampi1001, line 75: Small nuance worth considering (not blocking): the build job publishes whatever is on `main` without running the quality gate. Today that is safe because branch protection requires the `gate` check for PR merges — but a direct push to `main` by an admin bypasses branch protection, and this workflow would then publish an ungated build (CI runs in parallel but does not block this one). One cheap belt-and-braces option: add a `run: bash scripts/verify.sh --full` step (with `VITE_TMDB_READ_TOKEN` env) before Build here, mirroring `ci.yml`. If you prefer keeping deploys fast, skipping it is defensible — just noting the gap explicitly.

SrLampi1001 (approved):
Approving — clean, focused deploy pipeline that fully satisfies issue #18's acceptance criteria.

**What is solid**

- **Right-sized choice of target.** GitHub Pages via the official `configure-pages` / `upload-pages-artifact` / `deploy-pages` actions, with minimal permissions (`contents: read`, `pages: write`, `id-token: write`) and a `pages` concurrency group so overlapping pushes cannot race.
- **Both Pages quirks handled in the pipeline, not in app code.** Base path arrives through the new `CINETECA_BASE_PATH` env knob in `vite.config.ts` (default `/` untouched — dev server, CI gate and local preview are unaffected); SPA deep links survive hard reloads via the `index.html → 404.html` copy with no link rewriting or sessionStorage hacks.
- **Fails fast with an actionable error** when `VITE_TMDB_READ_TOKEN` is missing instead of shipping a build that crashes at runtime during env validation.
- **Consistent toolchain**: pnpm/Node 22/Corepack setup mirrors `ci.yml` exactly; no version drift between the two workflows.
- **README documents everything** (#18's "add the deploy URL to README" criterion): public URL, deep-link mechanism, one-time secret setup, and the correct note that a read-access token in a public bundle is by design.
- **TMDB attribution** is already rendered in the footer on every screen and covered by `root-layout.spec.tsx` / `router.spec.tsx`.

**Verified locally on `ea2dd96`:**

- Full gate green: `format:check`, `lint`, `check-types`, 452/452 tests.
- Built exactly like the workflow does (`CINETECA_BASE_PATH=/Project-save-me-of-the-riwi/ pnpm build` + `404.html` copy): every asset URL is base-prefixed, zero unprefixed references.
- Served `dist/` through `vite preview` and confirmed a hard request to `/Project-save-me-of-the-riwi/explore?genre=28` returns 200 with base-prefixed script tags.

**Two non-blocking nits (see line comments)**

1. The fail-fast secret check could run before `pnpm install` so the failure path doesn't pay the install cost.
2. The deploy build job publishes without running the quality gate — safe under branch protection, but an admin direct-push to `main` would publish ungated code. Consider mirroring `scripts/verify.sh --full` here if you want belt-and-braces.

One operational note for post-merge: the first workflow run needs the `VITE_TMDB_READ_TOKEN` repository secret to exist, otherwise it will fail fast with the message above — that's the intended behaviour per #18.

Issue #82, opened 2026-08-24
[Feature]: Wire the edit-list page and consume the reserved `lists.*` copy keys


SrLampi1001:
## Checklist
- [x] I have searched the existing issues

## Problem Statement

The local-lists feature shipped in #81 ships the create and detail surfaces, but the **edit-list surface is missing**. Several pieces are reserved for it and currently dangling:

- The form component (`src/presentation/components/feature/list-form.tsx`) accepts `mode: 'edit'`, but no route mounts the form with that mode.
- `useLists.update(list)` exists in `src/presentation/hooks/use-lists.ts` and `LocalListsRepository.update` exists, but no page consumes them.
- Five copy keys in `src/presentation/copy/strings.ts` were defined for the edit page and are now dead code:
  - `lists.editTitle` — "Edit list"
  - `lists.editDescription` — "Rename the list or update its description."
  - `lists.submitSuccessEdit` — "Changes saved"
  - `lists.removeConfirm` — "Delete this list?"
  - `lists.movieAlreadyInList` — "Already in this list"

Issue #16 ("Local lists (create, edit, validate)") explicitly lists **edit** as part of the scope. PR #81 only landed create + read + delete, so the edit path needs a follow-up.

## Proposed Solution

Wire the missing edit route and consume the reserved copy keys.

**Route**

- Add `/my-lists/:listId/edit` to `src/presentation/routes/router.tsx`.
- Render the existing `ListForm` with `mode="edit"` and the list's current values as `defaultValues`.
- On submit, call `useLists.update(...)`. On success, navigate back to `/my-lists/:listId`.

**Delete confirm**

- The "Delete list" button on `list-detail-page.tsx:170-178` is currently a one-click destructive action. While wiring edit, also add a confirmation step (a confirm dialog or `window.confirm` if the team accepts it) using the reserved `copy.lists.removeConfirm` key. This is also worth filing as a separate a11y issue if it grows in scope.

**Already-in-list hint**

- Use `copy.lists.movieAlreadyInList` on the detail page to mark movies that are already in the current list when the user opens it (the "Add to list" button currently toggles to "Remove from list" — a hint label clarifies the state).

## Alternatives Considered

- **Build edit and delete-confirm in two separate PRs.** That is reasonable if either grows non-trivial. The smallest viable follow-up is the edit route alone; delete-confirm is independently scoped.
[...]

PR #88, opened 2026-08-24
feat(infra): typed schema errors + 500-page clamp at the API boundary

### LINKED ISSUE #24: [Feature]: TMDB endpoint modules + Zod schemas

SebastianT2006:
## Description

Closes the two acceptance criteria of issue #24 that the earlier endpoint work (PRs #72/#77) left open. The five core endpoint modules, their Zod schemas, and the MSW handlers already exist on `main`; this PR hardens the boundary itself:

### 1. Drifted responses surface as a typed error, not a Zod dump

A response whose shape drifted (renamed field, string where a number belongs) used to throw a raw `ZodError` — a giant English parse dump with no context about which endpoint lied. Every module now parses through a shared `parseWith(schema, data, endpoint)` helper that translates the failure into **`TmdbSchemaError`**:

- names the lying endpoint (`/3/discover/movie`),
- carries one human-readable line per failing path (`issues: ['page: Expected number, received string', …]`),
- preserves the original `ZodError` as `cause`,
- mirrors the existing `TmdbHttpError` pattern from the network edge, and is exported from the api index so callers pattern-match one boundary error family.

Applied across all seven modules (`configuration`, `genres`, `discover`, `search`, `movie`, `recommendations`, `trending`) plus the two shared mappers.

### 2. Pagination respects the 500-page cap at this layer

The cap only lived in presentation hooks (`use-explore-filters`, `use-discover-movies`); the API layer forwarded any page and relied on TMDB rejecting it with body code 22. All four paginated modules (`discover`, `search`, `trending`, `recommendations`) now clamp the requested page into `[1, 500]` via a shared `clampPage` helper — `undefined`/non-finite stays unset, `< 1` → `1`, `> 500` → `500`, fractions truncated. Same belt-and-braces reasoning as the URL parser.

Closes #24

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [ ] Performance improvement
- [x] Test only _(17 new tests)_

## What Changed

- `src/infrastructure/api/_shared.ts` — new `TmdbSchemaError`, `parseWith`, `TMDB_MAX_PAGE`, `clampPage`; mappers switched to `parseWith`.
- `src/infrastructure/api/{discover,search,trending,recommendations}.ts` — `parseWith` + page clamp.
- `src/infrastructure/api/{configuration,genres}.ts` — `parseWith`.
- `src/infrastructure/api/index.ts` — exports `TmdbSchemaError`.
- `src/infrastructure/api/_shared.spec.ts` — _new_, 12 tests for the helpers.
- `discover/search/trending.spec.ts` — 5 integration tests: typed-error path via a broken MSW field, cap clamping observed on captured request URLs.

## Affected Layers

### Before / After

N/A — infrastructure behaviour, verified by tests.

## How to Test

1. Pull the branch: `git fetch && git checkout feature/infra-endpoint-hardening`
2. Run `pnpm test` — 449 total, all green; watch the new `_shared.spec.ts` plus the clamp/schema-error cases in discover/search/trending specs.
3. Quick manual probe of the typed error: point an MSW handler at `/3/discover/movie` returning `{ page: "oops", results: [] , ...}` and observe `TmdbSchemaError: TMDB response from /3/discover/movie failed validation (1 issue(s)): page: …`.

## Tests

- [x] I added unit tests for the new logic
- [x] I updated existing tests that no longer apply
- [x] Coverage has not decreased (domain still 100 % lines)

## Quality Gate

- [x] `pnpm build` succeeds

Verified locally with `bash scripts/verify.sh --full` (86 s, GREEN).

## Checklist

## Deployment Notes

- _None_ — no env vars, no migrations, no config toggles. Error-type swap is additive: anything that caught generic `Error` keeps working.

## Reviewer Focus

- @reviewer please look at
  - `src/infrastructure/api/_shared.ts` — `TmdbSchemaError` message construction (first issue inlined, full list in `.issues`) and `clampPage`'s truncation-before-clamp order.
  - `src/infrastructure/api/discover.spec.ts` — the structural assertion on the caught error (independent of how the class resolves through the spec's dynamic import).

### DIFF
--- a/src/infrastructure/api/_shared.ts
+++ b/src/infrastructure/api/_shared.ts
@@ -22,6 +22,85 @@ import { type RatingReliability } from '@/domain/rating/rating-reliability';
 import { type Trailer } from '@/domain/movie/trailer';
 
 // =============================================================================
+// Boundary errors + shared parsing
+// =============================================================================
+//
+// A broken field in a TMDB response must surface as a typed,
+// readable boundary error — never as a raw `ZodError` dump. This
+// mirrors the HTTP edge: network failures become `TmdbHttpError`
+// (see `infrastructure/http/errors.ts`), shape failures become
+// `TmdbSchemaError`.
+
+/**
+ * Thrown when a TMDB response does not match its documented shape.
+ * Carries the endpoint that lied plus a short list of human-readable
+ * issues so logs point at the actual drift instead of a parser dump.
+ */
+export class TmdbSchemaError extends Error {
+  override name = 'TmdbSchemaError' as const;
+  /** The endpoint whose response failed validation, e.g. `/3/discover/movie`. */
+  readonly endpoint: string;
+  /** Human-readable rendering of every failing path. */
+  readonly issues: readonly string[];
+
+  constructor(endpoint: string, cause: z.ZodError) {
+    const issues = cause.issues.map((issue) => {
+      const path = issue.path.length > 0 ? issue.path.join('.') : '<root>';
+      return `${path}: ${issue.message}`;
+    });
+    super(
+      `TMDB response from ${endpoint} failed validation (${String(issues.length)} issue(s)): ${issues[0] ?? 'unknown'}`,
+    );
+    this.endpoint = endpoint;
+    this.issues = issues;
+    this.cause = cause;
+  }
+}
+
+/**
+ * Parse untrusted data against a Zod schema at the API boundary.
+ * On failure, throws `TmdbSchemaError` (never a raw `ZodError`) so
+ * callers pattern-match one boundary error family, like they do
+ * for `TmdbHttpError`.
+ */
+export function parseWith<T>(schema: z.ZodType<T>, data: unknown, endpoint: string): T {
+  try {
+    return schema.parse(data);
+  } catch (cause: unknown) {
+    if (cause instanceof z.ZodError) {
+      throw new TmdbSchemaError(endpoint, cause);
+    }
+    throw cause;
+  }
+}
+
+// =============================================================================
+// Pagination guard
+// =============================================================================
+
+/**
+ * TMDB documents a hard cap of 500 pages on every paginated
+ * endpoint; asking beyond it returns TMDB body code 22. The API
+ * layer clamps instead of trusting the caller so the guarantee
+ * holds even if a presentation-layer guard is added later —
+ * the same belt-and-braces reasoning as the URL parser.
+ */
+export const TMDB_MAX_PAGE = 500;
+
+/**
+ * Clamp a requested page into `[1, TMDB_MAX_PAGE]`.
+ *
+ *   - `undefined` / non-finite → `undefined` (param not sent).
+ *   - `< 1`                    → `1`.
+ *   - `> 500`                  → `500`.
+ *   - fractional values        → truncated toward zero first.
+ */
+export function clampPage(page: number | undefined): number | undefined {
+  if (page === undefined || !Number.isFinite(page)) return undefined;
+  return Math.min(Math.max(Math.trunc(page), 1), TMDB_MAX_PAGE);
+}
+
+// =============================================================================
 // Raw TMDB Zod schemas
 // =============================================================================
 //
@@ -151,7 +230,7 @@ function toTrailers(rawArr: readonly z.infer<typeof tmdbTrailerSchema>[]): reado
  * `released` / `unreleased` boundary to a known instant.
  */
 export function toMovieSummary(raw: unknown, now: Date = new Date()): MovieSummary {
-  const parsed = tmdbMovieSummarySchema.parse(raw);
+  const parsed = parseWith(tmdbMovieSummarySchema, raw, 'shared/movie-summary');
   return {
     id: parsed.id,
     title: parsed.title,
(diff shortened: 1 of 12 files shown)


src/infrastructure/api/_shared.ts
```diff
@@ -151,7 +230,7 @@ function toTrailers(rawArr: readonly z.infer<typeof tmdbTrailerSchema>[]): reado
  * `released` / `unreleased` boundary to a known instant.
  */
 export function toMovieSummary(raw: unknown, now: Date = new Date()): MovieSummary {
-  const parsed = tmdbMovieSummarySchema.parse(raw);
+  const parsed = parseWith(tmdbMovieSummarySchema, raw, 'shared/movie-summary');
```
SrLampi1001, line 233: Nit (pre-existing, made more visible by this PR): every result item is validated twice per request — once inside each module's envelope schema (`results: z.array(tmdbMovieSummarySchema)`) and again here in `toMovieSummary` via `parseWith`. Harmless for correctness (the second parse always succeeds if the first did) but it doubles schema work on every 20-item page. Optional follow-up: have the envelope mappers accept the already-narrowed `z.infer<typeof tmdbMovieSummarySchema>` instead of `unknown`, or drop the item array from the envelope schemas and let the mapper be the single validator. Not blocking — fine to defer.

src/infrastructure/api/_shared.ts
```diff
@@ -170,7 +249,7 @@ export function toMovieSummary(raw: unknown, now: Date = new Date()): MovieSumma
  * domain `MovieDetail`. `now` is injectable for testing.
  */
 export function toMovieDetail(raw: unknown, now: Date = new Date()): MovieDetail {
-  const parsed = tmdbMovieDetailFullSchema.parse(raw);
+  const parsed = parseWith(tmdbMovieDetailFullSchema, raw, 'shared/movie-detail');
```
SrLampi1001, line 252: Small nuance on the "names the lying endpoint" guarantee: drift inside a detail body surfaces with `endpoint: 'shared/movie-detail'` rather than the actual URL (`/3/movie/{id}`), and item-level drift in discover/search/trending/recommendations reports `'shared/movie-summary'` — only the *envelope* parse names the real endpoint, while the more likely drift (a field inside a result) reports the mapper name. If you want the guarantee airtight, thread an optional endpoint label through `toMovieSummary` / `toMovieDetail` (each module already knows its URL) with the current strings as defaults. Defensible as-is since these are shared mappers; just flagging that the PR description's claim is only strictly true at the envelope level.

SrLampi1001 (approved):
Approving — this closes the two acceptance criteria of issue #24 that were still open, and does it with the right shape.

**What is solid**

- **`TmdbSchemaError` mirrors `TmdbHttpError` correctly**: endpoint name, one human-readable line per failing path in `.issues`, `<root>` rendering for pathless failures, original `ZodError` preserved as `cause`, and exported from the api index so callers pattern-match one boundary error family. The message inlines the first issue and keeps the full list on the error — good log ergonomics.
- **`parseWith` is applied everywhere it matters**: all seven modules (movie.ts transitively via the shared mappers) now throw typed errors, never raw Zod dumps. Non-Zod failures are rethrown untouched — correct restraint.
- **The clamp lives at the boundary now, not just in hooks**: all four paginated modules route through `clampPage`, which handles the full edge matrix (`undefined`/non-finite → unset, floor, cap, truncation-before-clamp). Keeping a private copy of the constant in presentation is fine here — presentation cannot import from infrastructure, so duplication across layers is the architecture working as designed.
- **Tests target behavior, not implementation**: the discover spec asserts structurally on the caught error (independent of class identity through dynamic imports), and clamp tests capture the outgoing request URL via MSW rather than trusting return values. 12 helper tests + 5 integration tests cover every branch of both helpers.

**Verified locally on `58c8268`:** format, lint, types, build, and dep check green; **469/469 tests pass**. One note for anyone reproducing: the `config/env > throws when VITE_TMDB_READ_TOKEN is missing` test fails if you have a local `.env.local` with a real token — environmental, not a defect; move the file aside to see the suite green.

**Two non-blocking nits (see line comments)**: (1) result items are schema-validated twice per request (envelope + mapper) — pre-existing, optional cleanup; (2) item-level drift reports `shared/movie-summary` as the endpoint instead of the real URL — only the envelope parse names the lying endpoint.

No new issues filed — nothing broken outside scope, and no overlap with open issues (#82/#83/#84/#86).

PR #81, opened 2026-08-24
feat(presentation): local lists (create, edit, validate, optimistic add/remove)

SebastianT2006:
## Description

Builds the themed-lists feature end-to-end across every hexagonal layer:

  - **domain** — `List` Zod schema, form schema with one Spanish message per
    rule, list factory (id + timestamps) the form does not own.
  - **application** — `ListsRepository` port with `create` / `update` /
    `remove` / `addMovie` / `removeMovie`, plus `ListNotFoundError`.
  - **infrastructure** — `LocalListsRepository` with corruption handling
    (log + reset, never silent) and `ListsCorruptedError`. Storage key
    `cineteca:lists:v1`. Singleton wired in `infrastructure/storage/index.ts`
    with `__resetListsRepositoryForTests` and `__setListsRepositoryForTests`.
  - **presentation** — `useLists` hook with TanStack Query optimistic
    mutations and rollback, three new routes (`/my-lists`,
    `/my-lists/new`, `/my-lists/:listId`), the `<ListForm>` wired through
    `react-hook-form` + `@hookform/resolvers/zod`, a `ListCard` for the
    overview, and a new "My Lists" entry in the primary nav.

### Form contract (issue #16 acceptance criteria)

- One Zod schema (`listFormSchema`) validates and types the form —
  `ListFormValues` is inferred from the schema via `z.infer`.
- Each rule has its own Spanish message; never a generic "can't do":
  - `El nombre es obligatorio.` (min 1)
  - `El nombre no puede tener más de 50 caracteres.` (max 50)
  - `La descripción no puede tener más de 20 caracteres.` (max 20)
- Errors are announced to screen readers via `role="alert"` regions;
  each field carries `aria-invalid` and `aria-describedby` pointing at
  the error message.
- Focus moves to the first invalid field on submit failure through
  react-hook-form's `setFocus`.
- Submit is blocked while the mutation is in flight (`isBusy` disables
  every interactive control and switches the submit label to
  "Saving…").
- The success state is announced via `role="status"` + `aria-live="polite"`.
- Tests are written with `getByRole` / `getByLabelText` / `getByText` —
  no `data-testid` hooks on the form.

### Detail page contract

- Movies are joined against the local library so a single `MovieCard`
  per row is enough (library is the source of truth for movie metadata).
- `addMovie` / `removeMovie` use the optimistic-mutation hook, so the
  UI reflects the new state before the storage write returns.
- The page renders a "list not found" empty state when the id does not
[...]

PR #89, opened 2026-08-24
docs(github): add branch naming reminder to PR template

3105jero:
## Description

Closes #28

The branching model is already documented in `CONTRIBUTING.md` (long-lived + short-lived branch tables, naming rules, examples, and merge strategy). What is missing is a *reminder* at the top of every PR so authors self-check the branch prefix, the kebab-case shape, and the correct base before requesting review.

The other DoD item from #28 — *"`develop` branch is created off `main`"* — is satisfied out of band by this PR's setup: `develop` was created at `9175135` (the tip of `main` at the time) and is now the base for this and every future PR.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional change, just code improvement)
- [x] Chore / Maintenance (dependency upgrade, tooling, docs)
- [ ] Performance improvement
- [ ] Test only

## Branch Naming

- [x] My branch starts with `feature/`, `bugfix/`, `chore/`, `release/`, or `hotfix/`
- [x] My branch name is lowercase, kebab-case, and follows `<prefix>/<scope>-<short-desc>` (no issue number in the name)
- [x] I branched off `develop` (or off `main` for `hotfix/` and `release/`)

## What Changed

- `.github/pull_request_template.md` — new **Branch Naming** section between **Type of Change** and **What Changed**, with three checklist items (prefix, kebab-case shape, correct base) and a comment pointing at `CONTRIBUTING.md`.

## Affected Layers

- [ ] `domain/` — pure TypeScript, no framework imports
- [ ] `application/` — use cases and ports
- [ ] `infrastructure/` — HTTP, storage, API
- [ ] `presentation/` — React, routes, hooks

## How to Test

1. Pull `chore/docs-add-branch-prefix-reminder`
2. Open any test PR against `develop`
3. Confirm the **Branch Naming** section renders with the three checkboxes
4. Confirm the existing sections (Description, Type of Change, What Changed, …) still render in the same order

## Tests

- [ ] I added unit tests for the new logic (N/A — docs change)
- [x] I updated existing tests that no longer apply (none)
- [ ] I verified the manual test plan above (visual only)
- [x] Coverage has not decreased (no source touched)

## Quality Gate

- [x] `pnpm format:check` passes (Prettier, no changes)
- [x] `pnpm lint` passes (zero warnings)
[...]

PR #90, opened 2026-08-24
fix(test): restore jsdom localStorage under Node v22+ built-in stub

3105jero:
## Description

Closes the failing-test cascade introduced by Node v22+'s built-in `localStorage` global. The change is **18 lines in `vitest.setup.ts`**, with no source-code or spec changes.

### Background

Node v22+ exposes a `localStorage` global as part of the Web Storage API. When `--localstorage-file` is not provided to a valid path (the default for `pnpm test`), the global is an **empty stub object with no methods** (`setItem`/`getItem`/`clear`/etc. are all `undefined`).

Vitest 4's `populateGlobal` walks the jsdom window and only assigns keys that are NOT already on `global` (or are in its allow-list). Since `localStorage` is neither, vitest skips it — and Node's empty stub wins over jsdom's real `Storage`.

The `afterEach` in `vitest.setup.ts:48` calls `window.localStorage.clear()` after every test. When that throws, three cleanup steps silently break:
1. `cleanup()` from `@testing-library/react` does not run → DOM is not reset.
2. `__resetLibraryRepositoryForTests()` / `__resetListsRepositoryForTests()` do not run → the singleton repositories keep state from the previous test.
3. The next test sees dirty DOM, dirty storage, and mutated repositories, and fails for a different reason.

The cascade explains the **590 failures** seen on `pnpm test`: most are not bugs in the specs, they are the afterEach teardown failing on the very first test and contaminating every subsequent one. Isolating a single spec file (`save-to-library-button.spec.tsx`) shows only 6 failures — the real ones.

### Fix

In `vitest.setup.ts`'s `beforeAll`, before MSW listens, replace `globalThis.localStorage` with jsdom's real `Storage` instance via `Object.defineProperty`. Guarded by `if (jsdomInstance)` so non-jsdom environments (e.g. edge-runtime if added later) are untouched.

```ts
const jsdomInstance = (globalThis as { jsdom?: { window: { localStorage: Storage } } }).jsdom;
if (jsdomInstance) {
  Object.defineProperty(globalThis, 'localStorage', {
    value: jsdomInstance.window.localStorage,
    writable: true,
    configurable: true,
  });
}
```

`globalThis.jsdom` is the handle vitest exposes on the worker global **just before** setupFiles run, so the override is in place before any spec executes.

### Verification

| Check | Before fix | After fix |
| --- | --- | --- |
| `pnpm test` (full) | 590 failed | **469 passed** in 12.94s |
| Coverage statements | (skipped — coverage gate blocked) | 91.3% (≥80% ✅) |
| Coverage branches | — | 84.62% (≥80% ✅) |
| Coverage `src/domain/**` | — | 100% (≥100% ✅) |
| `pnpm format:check` | ✅ | ✅ |
| `pnpm lint` | ✅ | ✅ |
| `pnpm check-types` | ✅ | ✅ |
| `pnpm build` | ✅ | ✅ |

### Notes

- The fix is a **runtime workaround**, not a version pin. If/when vitest adds `localStorage` to its allow-list (or jsdom stops honouring Node's stub), the override becomes a no-op and can be deleted.
- One concern per branch: no source-code or spec changes ride along.

Closes #81 (in part — the spec files added in that PR exercised `window.localStorage` and unmasked the bug; this fix unblocks them).

## Type of Change

- [ ] Performance improvement
- [ ] Test only

## Branch Naming

- [x] My branch starts with `feature/`, `bugfix/`, `chore/`, `release/`, or `hotfix/` (`fix/` — accepted shorthand for `fix(test):`)
- [x] My branch name is lowercase, kebab-case, and follows `<prefix>/<scope>-<short-desc>` (no issue number in the name)
- [x] I branched off `develop` (or off `main` for `hotfix/` and `release/`)

> Side note: the branch is `fix/...` rather than `chore/...` because `chore/` is reserved in this repo for tooling that does not change behaviour, and this PR **does** change behaviour (restores the failing teardown). Conventional Commits allows `fix:` for any bug, including infrastructure bugs.

## What Changed

- `vitest.setup.ts` — added 18 lines in `beforeAll` that override `globalThis.localStorage` with the real `Storage` exposed by jsdom. Guarded by `if (jsdomInstance)` so the override is a no-op on non-jsdom environments.

## Affected Layers

## How to Test

1. Pull `fix/test-restore-jsdom-localstorage`
2. Run `bash scripts/verify.sh --full`
3. Confirm `Tests  469 passed (469)` and no `Test Files  failed`
4. (Regression check) `pnpm exec vitest run --no-coverage src/presentation/components/feature/save-to-library-button.spec.tsx` — must show 6/6 pass

## Tests

- [ ] I added unit tests for the new logic (N/A — runtime test-environment fix; behaviour is covered by the 469 specs that now pass)
- [ ] I updated existing tests that no longer apply (none)
- [ ] I verified the manual test plan above (469/469 + 6/6 regression spot-check)
- [x] Coverage has not decreased (91.3% statements, 84.62% branches, 89.53% functions, 93.06% lines; `src/domain/**` still 100%)

## Quality Gate

- [x] `pnpm build` succeeds
- [x] `./scripts/check-versions.sh` reports no deprecated deps introduced by this PR

## Checklist

- [x] I have added comments only where the logic is non-obvious (block comment explains the Node-vs-vitest interaction)
- [x] I have updated the documentation (README, copy, etc.) if needed (N/A — runtime workaround; comment in source is sufficient)

## Deployment Notes

- _None_ — test-environment only; production bundles are unaffected.

## Reviewer Focus

- `@reviewer` please look at `vitest.setup.ts` lines 16–33 — confirm the `Object.defineProperty` semantics are correct (`writable: true` lets future test code mutate `localStorage`; `configurable: true` lets the override be torn down if needed). The cast `globalThis as { jsdom?: ... }` is intentional because `jsdom` is a vitest-only runtime handle and is not in the project's ambient types.
- Please also confirm the branch prefix is acceptable. `fix/` is not in the documented allow-list (`feature/`, `bugfix/`, `chore/`, `release/`, `hotfix/`); if the reviewer prefers `bugfix/test-restore-jsdom-localstorage` or `chore/test-restore-jsdom-localstorage`, say the word and I'll rename + force-push the branch.

---

### DIFF
--- a/vitest.setup.ts
+++ b/vitest.setup.ts
@@ -13,6 +13,24 @@ vi.stubEnv('VITE_TMDB_READ_TOKEN', 'a'.repeat(64));
 // `onUnhandledRequest: 'error'` is what makes MSW useful: a request that
 // nobody simulated blows up the test instead of going to the real network.
 beforeAll(() => {
+  // Node v22+ ships a built-in `localStorage` global as part of the
+  // Web Storage API. When `--localstorage-file` is not provided to a
+  // valid path, the global is an empty stub object with no methods
+  // (`setItem`/`getItem`/`clear`/etc. are all `undefined`). Vitest's
+  // `populateGlobal` walks the jsdom window and only assigns keys that
+  // are NOT already on `global` (or are in its allow-list); since
+  // `localStorage` is neither, vitest skips it and Node's stub wins.
+  // Every spec that touches `window.localStorage` then explodes with
+  // "localStorage.clear is not a function". Replacing the stub with
+  // jsdom's real Storage at file setup restores the pre-v22 behaviour.
+  const jsdomInstance = (globalThis as { jsdom?: { window: { localStorage: Storage } } }).jsdom;
+  if (jsdomInstance) {
+    Object.defineProperty(globalThis, 'localStorage', {
+      value: jsdomInstance.window.localStorage,
+      writable: true,
+      configurable: true,
+    });
+  }
   server.listen({ onUnhandledRequest: 'error' });
 });
 afterEach(async () => {


vitest.setup.ts
```diff
@@ -13,6 +13,24 @@ vi.stubEnv('VITE_TMDB_READ_TOKEN', 'a'.repeat(64));
 // `onUnhandledRequest: 'error'` is what makes MSW useful: a request that
 // nobody simulated blows up the test instead of going to the real network.
```
SebastianT2006, line 14: The mechanism described here is real and I verified the fix works — but the version claim is wrong, and this comment block (plus the PR body) bakes it in for future readers.

Per [Node's official globals docs](https://nodejs.org/api/globals.html#localstorage):

- **v22.4.0**: `localStorage` added — but only behind `--experimental-webstorage`. On stock Node 22 (what CI pins) it does not exist at all; I confirmed `typeof localStorage === 'undefined'` on v22.23.2.
- **v25.0.0**: unflagged (enabled by default). Without `--localstorage-file`, accessing `localStorage` returns an **empty object** — exactly the stub described here.
- **v26**: accessing it without a file throws a `DOMException`.

So the failure hits contributors running local Node ≥ 25 while CI (Node 22) stays green — which is presumably why the author saw 590 failures locally and CI never flagged it.

Please reword "Node v22+" → "Node 25+ (unflagged in v25; v22 only exposes it behind `--experimental-webstorage`)" in both this comment and the PR description, so someone debugging on Node 22 knows why they can't reproduce.

vitest.setup.ts
```diff
@@ -13,6 +13,24 @@ vi.stubEnv('VITE_TMDB_READ_TOKEN', 'a'.repeat(64));
...
 beforeAll(() => {
+  // Node v22+ ships a built-in `localStorage` global as part of the
+  // Web Storage API. When `--localstorage-file` is not provided to a
+  // valid path, the global is an empty stub object with no methods
+  // (`setItem`/`getItem`/`clear`/etc. are all `undefined`). Vitest's
+  // `populateGlobal` walks the jsdom window and only assigns keys that
+  // are NOT already on `global` (or are in its allow-list); since
+  // `localStorage` is neither, vitest skips it and Node's stub wins.
+  // Every spec that touches `window.localStorage` then explodes with
+  // "localStorage.clear is not a function". Replacing the stub with
```
SebastianT2006, line 24: Verified this empirically — the fix is correct:

1. Simulated a Node 25 environment (`--require` preload defining `globalThis.localStorage = {}`, matching v25's documented "returns an empty object" behaviour) and ran `develop` without this patch: `save-to-library-button.spec.tsx` fails 6/6 with exactly the described cascade.
2. Same simulation against this branch: 6/6 pass, and the full suite is **469/469** — no contamination between tests, so the `afterEach` teardown (cleanup + repo resets + `localStorage.clear()`) runs again.
3. Without the simulation (Node 22.23.2), this branch stays fully green: 469/469, `eslint --max-warnings 0`, `tsc --noEmit`. When Node does not define the global, vitest's jsdom `populateGlobal` has already installed the real Storage and the `defineProperty` re-asserts the same value — harmless no-op.

The guard on vitest's `globalThis.jsdom` handle and the `writable/configurable` flags are fine. Nice that it degrades to a no-op if upstream fixes the key assignment.

SebastianT2006 (changes requested):
Solid catch and a correct, well-isolated fix — I reproduced the failure mode and verified the patch resolves it (evidence in the inline comments). However, I'm requesting changes on four points; all are quick:

**1. Wrong version claim: "Node v22+" → it's Node 25+** (see inline comment). Stock Node 22 doesn't even define `localStorage`; the empty-stub behaviour is v25.0.0+ ([Node globals docs](https://nodejs.org/api/globals.html#localstorage)). This explains why CI is green while local runs broke. Please fix the wording in `vitest.setup.ts` and in the PR description.

**2. Branch prefix `fix/` is not in the allow-list.** The template you just helped land (#89) and `CONTRIBUTING.md` allow only `feature/`, `bugfix/`, `chore/`, `release/`, `hotfix/`, and rule 1 says mismatched branches are sent back. You offered it yourself in Reviewer Focus: please rename to `bugfix/test-restore-jsdom-localstorage`. Your reasoning for *not* using `chore/` is sound — this changes behaviour, so `bugfix/` is the right home.

**3. "Closes #81" is not actionable.** #81 is a **pull request** (the local-lists feature), already merged — there is nothing to close, and `Closes` wouldn't fire from `develop` anyway. Use `Refs #81` to credit it as the change that unmasked the bug.

**4. Type of Change should be "Bug fix", not "Chore / Maintenance".** By your own (correct) argument that this restores failing behaviour, the checkbox and title type agree with each other (`fix(test)`) but not with the checked box.

Minor, non-blocking: scope `test` isn't in the allowed scopes list (`gate` covers gate/tooling-adjacent areas) — fine to keep if the team prefers, but worth aligning eventually.

Non-blocking observation filed separately: the root cause here is toolchain drift — CI pins Node 22, contributors run whatever is installed locally, and the repo documents no supported range (no `engines`, no `.nvmrc`). Tracking that in its own issue so the fix isn't only a workaround in test setup.

3105jero:
Closing in favour of #92 — same fix, renamed `fix/` → `bugfix/` per your branch-prefix review and corrected the Node v22+ → Node 25+ version claim. Body of #92 has the full diff context.

Will delete the `fix/test-restore-jsdom-localstorage` branch in a follow-up commit.

PR #98, opened 2026-08-25
fix(infra): stop requiring genre_ids in /movie/{id} schema

3105jero:
## Description

Fixes the `MovieDetailPage` rendering its error state for every movie. The page looked like a 404 to the user, but TMDB itself was returning **200** with the real detail body — the failure was on the consumer side.

### Root cause

`tmdbMovieDetailFullSchema` in `src/infrastructure/api/_shared.ts` was copied from the summary schema and required `genre_ids: z.array(z.number())`. TMDB's `/movie/{id}` endpoint **does not return `genre_ids`** — it returns full `genres: [{id, name}]` objects. The numeric `genre_ids` array is reserved for the summary endpoints (`/discover/movie`, `/search/movie`, `/trending`, `/recommendations`).

Result of the mismatch: `parseWith(tmdbMovieDetailFullSchema, raw, ...)` threw `TmdbSchemaError` on every real TMDB response, `useMovieDetail` rejected, and the page rendered its error state — whose `<h1>` says `copy.detail.notFoundTitle` ("Película no encontrada"), so the user saw what looked like a 404.

### Why the test suite missed it

`src/infrastructure/api/movie.spec.ts` used a fabricated `detailFixture` with `genre_ids: [28, 12]` injected — TMDB never returns that field on this endpoint, so the tests were validating an invented response shape, not the real one. The "throws when a required field is missing" test was even more confused: it had a fixture with `genre_ids: []` so the schema pass was synthetic, and the test name no longer matched what it actually asserted.

### Fix

- `src/infrastructure/api/_shared.ts` — drop `genre_ids` from `tmdbMovieDetailFullSchema`; document the why in a schema-level comment so the next contributor doesn't re-add it. In `toMovieDetail`, derive `MovieDetail.genreIds` from `parsed.genres.map(g => g.id)` instead.
- `src/infrastructure/api/movie.spec.ts` — drop the fabricated `genre_ids` from `detailFixture` so it matches the real TMDB shape; add a regression assertion `expect(detail.genreIds).toEqual([28, 12])` whose comment names the symptom (page shows the error state) so the failure mode is obvious if someone reintroduces the schema field.

### Behaviour after the fix

Open `/movie/{id}` for any real movie → page renders the movie, not the error state. `MovieDetail.genreIds` is populated from the full `genres` objects (same data, different endpoint).

### Known UX limitation, out of scope here

The error state in `movie-detail-page.tsx:114` renders `copy.detail.notFoundTitle` in the `<h1>` for *every* error kind (schema, network, server). After this fix the only realistic triggers are real TMDB outages or `notFound` — but if a future schema or transport error surfaces, the page will still look like a 404 to the user. One-line follow-up: switch the `<h1>` to `copy.errors.title` when `detail.error.detail.kind !== 'notFound'`. Flagging here so the next contributor doesn't repeat the diagnosis; not bundled with this fix to keep the PR single-concern.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Performance improvement
- [ ] Test only

## Branch Naming

- [x] My branch starts with `feature/`, `bugfix/`, `chore/`, `release/`, or `hotfix/`
- [x] My branch name is lowercase, kebab-case, and follows `<prefix>/<scope>-<short-desc>` (no issue number in the name)
- [x] I branched off `develop`

## What Changed

- `src/infrastructure/api/_shared.ts` — `tmdbMovieDetailFullSchema` no longer requires `genre_ids`; added a schema-level doc comment explaining why. `toMovieDetail` now derives `genreIds` from `parsed.genres.map((g) => g.id)`.
- `src/infrastructure/api/movie.spec.ts` — removed fabricated `genre_ids` from `detailFixture` (the old fixture reflected an invented TMDB shape, not the real one). Added regression assertion `expect(detail.genreIds).toEqual([28, 12])` with a comment naming the user-visible symptom if the schema is ever reverted.

## Affected Layers

### Before

_App shows the error-state `<h1>` "Película no encontrada" with the retry CTA. Terminal shows TMDB returning 200. Symptom is the wrong heading + retry CTA on every detail page._

### After

_App shows the movie hero, meta panel, cast, trailers, and recommendation rail as designed. No regression possible for the `genreIds` consumer (`SaveToLibraryButton` builds a `LibraryEntry` from `movie.genreIds` at `movie-detail-page.tsx:189` — same array, just sourced from `genres[].id` instead of the absent `genre_ids`)._

## How to Test

1. Pull `bugfix/infra-movie-detail-genre-ids`
2. `pnpm install` (no new deps; lockfile unchanged)
3. `pnpm dev` and visit `/movie/27205` (Inception, fixture matches this in `movie.spec.ts`) — page should render the hero, not the error state
4. Open DevTools Network → confirm the request to `/3/movie/27205?append_to_response=credits,videos` returns 200 and the parsed `genres[]` is non-empty
5. Visit any other real `/movie/{id}` (e.g. `/movie/550` Fight Club) — page renders normally
6. Run `pnpm exec vitest run src/infrastructure/api/movie.spec.ts` — all 7 tests pass; the regression assertion at the bottom of "fetches the detail and maps the basic fields to MovieDetail" is the tripwire if anyone re-adds `genre_ids` to the schema

## Tests

- [x] I added unit tests for the new logic (regression assertion that `MovieDetail.genreIds` is derived from `genres[]` and would fail if `genre_ids` is reintroduced to the schema)
- [x] I updated existing tests that no longer apply (removed fabricated `genre_ids` from `detailFixture` and from the "throws when a required field is missing" fixture)
- [x] I verified the manual test plan above (full gate green; manual page render verified against the `27205` Inception fixture)
- [x] Coverage has not decreased (91.3% statements / 84.62% branches — same as `develop`)

## Quality Gate

- [x] `pnpm build` succeeds

Verified locally with `bash scripts/verify.sh --full` (89s, GREEN). Husky pre-commit + pre-push ran the same gate and passed.

## Checklist

- [x] My code follows the project's architecture rules (domain is pure, dependencies point inward — change is in `infrastructure/api/`, no layer crossing)
- [x] I have used the codebase's existing patterns and utilities (kept the `parseWith` / `TmdbSchemaError` boundary translation; no new abstractions)
- [x] I have added comments only where the logic is non-obvious (schema-level comment + mapper-side comment + test-side comment, each naming the user-visible failure mode)
- [x] I have updated the documentation (README, copy, etc.) if needed (N/A — runtime fix; no public docs reference the schema shape)

## Deployment Notes

_None_ — pure runtime fix; no env vars, migrations, config toggles, or breaking changes. Production bundle size unchanged (the schema lost one field, the mapper gained one `.map()` call).

## Reviewer Focus

- `@SebastianT2006` please look at `src/infrastructure/api/_shared.ts:166-182` — confirm the schema-level doc comment captures the why (so a future contributor doesn't re-add `genre_ids` from the summary schema).
- Please also confirm `src/infrastructure/api/movie.spec.ts:67-95` — the regression assertion at line ~93 is the tripwire; if anyone re-adds `genre_ids` to the detail schema, `getMovieDetail` will throw `TmdbSchemaError` and the page starts rendering its error state for every movie again.
- The "Known UX limitation" flagged in the Description (error-state `<h1>` shows `notFoundTitle` for every error kind) is intentionally **not** in this PR. If you want it bundled, say the word and I'll add the one-line change in a follow-up commit on the same branch.

---

### DIFF
--- a/src/infrastructure/api/_shared.ts
+++ b/src/infrastructure/api/_shared.ts
@@ -153,6 +153,16 @@ const tmdbDetailAppendagesSchema = z.object({
     .optional(),
 });
 
+/**
+ * Schema for the raw `/movie/{id}` TMDB response.
+ *
+ * Note: this endpoint does NOT include `genre_ids` (the summary
+ * endpoints do). The detail endpoint returns `genres: [{id, name}]`
+ * and the mapper derives `genreIds` from those. An earlier draft of
+ * this schema copied `genre_ids` from the summary schema, which made
+ * the real response fail validation and the page render its error
+ * state instead of the movie.
+ */
 export const tmdbMovieDetailFullSchema = z.object({
   id: z.number(),
   title: z.string(),
@@ -162,7 +172,6 @@ export const tmdbMovieDetailFullSchema = z.object({
   release_date: z.union([z.string(), z.null()]),
   vote_average: z.number(),
   vote_count: z.number(),
-  genre_ids: z.array(z.number()),
   tagline: z.union([z.string(), z.null()]),
   runtime: z.union([z.number().nonnegative(), z.null()]),
   genres: z.array(tmdbGenreSchema),
@@ -268,7 +277,7 @@ export function toMovieDetail(raw: unknown, now: Date = new Date()): MovieDetail
     backdropPath: noDataFromNullableString(parsed.backdrop_path),
     state: classifyRelease(parsed.release_date, now),
     rating: toRating(parsed.vote_count, parsed.vote_average),
-    genreIds: parsed.genre_ids,
+    genreIds: parsed.genres.map((g) => g.id),
     tagline: noDataFromNullableString(parsed.tagline),
     runtime: noDataFromNullableNumber(parsed.runtime),
     genres: toGenres(parsed.genres),
--- a/src/infrastructure/api/movie.spec.ts
+++ b/src/infrastructure/api/movie.spec.ts
@@ -5,6 +5,11 @@ import { server } from '@/test/msw/server';
 const BASE = 'https://api.tmdb.test';
 const TOKEN = 'a'.repeat(64);
 
+// Real TMDB `/movie/{id}` shape: this endpoint returns full `genres`
+// objects but no `genre_ids` (that field is reserved for the
+// summary endpoints — discover, search, trending, recommendations).
+// An earlier fixture fabricated `genre_ids` to satisfy the schema,
+// which masked the schema-vs-response drift that broke the page.
 const detailFixture = {
   id: 27205,
   title: 'Inception',
@@ -14,7 +19,6 @@ const detailFixture = {
   release_date: '2010-07-16',
   vote_average: 8.4,
   vote_count: 30_000,
-  genre_ids: [28, 12],
   tagline: 'Your mind is the scene of the crime.',
   runtime: 148,
   genres: [
@@ -80,6 +84,12 @@ describe('infrastructure/api/movie', () => {
       { id: 28, name: 'Action' },
       { id: 12, name: 'Adventure' },
     ]);
+    // Regression: `/movie/{id}` does not return `genre_ids`. The
+    // mapper derives `genreIds` from the `genres[]` objects. If the
+    // schema starts requiring `genre_ids` again, this assertion will
+    // fail (the call above will reject with TmdbSchemaError) and the
+    // page will start rendering its error state for every movie.
+    expect(detail.genreIds).toEqual([28, 12]);
     // Without the append flag, cast and trailers are absent.
     expect(detail.cast.kind).toBe('absent');
     expect(detail.trailers.kind).toBe('absent');
@@ -198,7 +208,6 @@ describe('infrastructure/api/movie', () => {
           genres: [],
           budget: 0,
           revenue: 0,
-          genre_ids: [],
         }),
       ),
     );


SrLampi1001 (approved):
Approving — root cause verified against the live API and the fix is minimal, correct, and well-guarded.

**Root cause independently confirmed:** hit the real `/3/movie/27205` with our read token — response contains **no** `genre_ids` key and returns full `genres` objects (`Action`, `Science Fiction`, `Adventure`). The schema was requiring a field TMDB never sends on this endpoint, so every real detail fetch threw `TmdbSchemaError` and the page rendered its "Película no encontrada" error state. Diagnosis in the PR body is exactly right, including why the suite missed it (fixture injected the fabricated field).

**What is solid**

- Fix at the correct layer: schema drops the impossible requirement; mapper derives `genreIds` from `genres[].id` — same domain data, sourced from where this endpoint actually provides it. No layer crossings, no new abstractions.
- Schema-level doc comment names both the endpoint asymmetry and the user-visible failure mode of re-adding it — exactly the tripwire documentation this kind of copy-paste drift needs.
- Fixture now matches reality; regression assertion `expect(detail.genreIds).toEqual([28, 12])` fails loudly if anyone reintroduces the field.
- Honest scoping: the `notFoundTitle`-for-every-error-kind UX limitation is correctly flagged as out-of-scope rather than smuggled in.

**Verified locally on `324fd54`:** format/lint/types green, build clean, 469/469 tests (with real `.env.local` present). A browser wasn't needed — the live-API probe plus MSW-backed specs cover both halves of the verification.

**One non-blocking follow-up (see line comment):** the shared MSW fixtures in `tmdb-handlers.ts` still carry the same fabricated shape — `TMDB_MOVIE_DETAIL_FIXTURE` spreads `genre_ids` from the summary fixture into a *detail* fixture under a header claiming it mirrors what TMDB actually returns. Inert today (Zod strips unknown keys) but it's the second copy of the exact invented shape this PR removes; worth a 2-line cleanup soon.

SrLampi1001 (commented):
Non-blocking follow-up noted outside the diff: `src/test/msw/tmdb-handlers.ts:51` — `TMDB_MOVIE_DETAIL_FIXTURE` spreads `TMDB_MOVIE_SUMMARY_FIXTURE`, dragging the fabricated `genre_ids` into a `/movie/{id}` detail fixture under a file header claiming these "mirror the shape TMDB actually returns" (they don't for this endpoint — verified live). Inert today since Zod strips unknown keys, but it's the second copy of the exact invented shape this PR removes; if the schema ever goes `.strict()` or someone asserts on fixture shape, presentation specs would validate an invented response again. Suggested 2-line cleanup: build the detail fixture independently (or destructure `genre_ids` out of the spread). Fine to do as a fast-follow.

PR #100, opened 2026-08-25
fix(types): unblock #62 by fixing the 8 type errors the broken gate hid

3105jero:
## Description

First of two batches preparing for issue #62 (`chore(gate): fix pnpm check-types no-op`). This PR fixes 8 of the 28 type errors that the broken `pnpm check-types` has been hiding — all four errors that the gate would have caught the moment `tsc -b --noEmit` becomes the default. The remaining 20 are big enough to warrant their own PR (rating rendering) or different category fixes (Axios error typing, `pickUnit` return type, `list-detail-page` narrowing) — they'll land in batch 2.

### Why this is a separate PR, not bundled with the gate flip

The gate flip (the one-line `tsc -b --noEmit` change in `package.json`) is the actual fix for #62. But flipping it without first fixing the 28 latent errors turns CI permanently red. So the sequencing has to be:

1. **This PR (batch 1)** — fix the 8 type errors that are cheap and well-understood.
2. **Batch 2 (next PR)** — fix the remaining 20: rating rendering (20), duration, Axios client, list-detail narrowing, spec narrowing.
3. **Batch 3 (last PR)** — flip the script in `package.json` to `tsc -b --noEmit`. The gate then becomes honest.

The gate stays green between every PR because the broken `pnpm check-types` is still in effect — `tsc -b --noEmit` is run manually in this PR's verification (and documented below) to prove the fixes are real, not a false-green.

### The four fixes

| Commit | Layer | Files | Errors fixed |
|---|---|---|---|
| `5bdf55e` | infra | `src/infrastructure/api/discover.ts` | 5 — `DiscoverSortOption` was missing `'title.asc'` / `'title.desc'` even though the URL parser's `SORT_VALUES` table and the Explore UI both use them |
| `b872a36` | application | `src/application/ports/library-repository.ts`, `src/application/ports/lists-repository.ts` | 2 — `Error.cause` is a built-in (ES2022); the `LibraryCorruptedError` and `ListsCorruptedError` subclasses need the `override` modifier under `noImplicitOverride` |
| `e85fcd3` | presentation | `src/presentation/routes/movie-detail-page.tsx` | 1 — `movie.genreIds` is `readonly number[]` (from `MovieSummary`); `LibraryEntry.genreIds` is mutable. Spread into a fresh array. **This was a regression I introduced in PR #98** and the broken gate hid it. |

### Verification that the fixes are real

Manual run on this branch:

```bash
npx tsc -b --noEmit --force
```

| Stage | Errors |
|---|---|
[...]

PR #99, opened 2026-08-25
feat(presentation): wire edit-list page and guard delete with a confirm dialog

SrLampi1001:
## Description

Solves #82 and #83 together (both touch `list-detail-page.tsx` and share the same copy keys, per #83's bundling note).

### 1. Edit-list surface (`/my-lists/:listId/edit`) — issue #82

- New `ListEditPage` renders the existing `ListForm` with `mode="edit"`, **pre-populated** with the list's current name/description.
- Submit calls `useLists.update(...)`. The repository stamps `updatedAt`; the page only sends name + description. A no-op result (list deleted in another tab) surfaces as an error instead of navigating.
- Success navigates back to `/my-lists/:listId`, which now shows an **Edit** link in its header (the route was unreachable otherwise).
- Unknown id renders the same not-found empty state as the detail page. Loading branch mirrors it too.

### 2. Delete confirmation — issue #83 (Option A: dialog)

- New `ConfirmDialog` component (`src/presentation/components/ui/confirm-dialog.tsx`):
  - `role="alertdialog"` + `aria-modal`, labelled by title, described by body.
  - **Initial focus on Cancel** (safe default for destructive actions).
  - Focus trap on Tab / Shift+Tab; Escape closes without confirming; focus restored to the trigger on close.
- The detail page's "Delete list" button now only opens the dialog; `lists.remove(id)` fires exclusively from the dialog's confirm button, which carries `aria-label="Delete list: {name}"`.
- Body combines the reserved `copy.lists.removeConfirm` with a new "This cannot be undone." line.

### 3. Reserved copy keys consumed

Re-adds the keys dropped in `92a8f52` and consumes all of them:

| Key | Consumed by |
| --- | --- |
| `lists.editTitle` | edit page heading + detail-page Edit link + document title |
| `lists.editDescription` | edit page header |
| `lists.submitSuccessEdit` | edit success announcement |
| `lists.removeConfirm` | dialog body |
| `lists.movieAlreadyInList` | hint under every in-list card |

New keys: `editAria` (form aria-label now follows mode), `errorEdit`, `deleteDialogTitle`, `deleteDialogBody`.

Closes #82
Closes #83

## Type of Change

- [ ] Bug fix
- [x] New feature (non-breaking)
- [ ] Breaking change
- [ ] Refactor
- [ ] Chore
- [ ] Performance
- [ ] Test only

## What Changed

- `src/presentation/routes/list-edit-page.tsx` — new edit screen.
[...]

PR #101, opened 2026-08-25
chore(test): keep the MSW detail fixture true to /movie/{id}

SrLampi1001:
## Description

Follow-up from the #98 review: `TMDB_MOVIE_DETAIL_FIXTURE` was built by spreading `TMDB_MOVIE_SUMMARY_FIXTURE`, which dragged the numeric `genre_ids` array into a fixture served for `/movie/{id}` — a field that endpoint never returns (verified live in the #98 review). The file header claims these fixtures "mirror the shape TMDB actually returns"; now they actually do, per endpoint.

The change is inert today (Zod strips unknown keys), but it removes the last copy of the invented shape so a future `.strict()` schema or fixture-shape assertion cannot silently validate an invented response again.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [x] Chore / Maintenance (test tooling only)
- [ ] Performance
- [ ] Test only

## What Changed

- `src/test/msw/tmdb-handlers.ts` — the detail fixture now omits `genre_ids` (destructuring-rest on the spread copy), with a comment naming the endpoint asymmetry and PR #98.

## Affected Layers

- [ ] `domain/`
- [ ] `application/`
- [ ] `infrastructure/`
- [ ] `presentation/`
- [x] `test/`

## How to Test

1. Pull `chore/test-msw-detail-fixture-shape`
2. `grep -n "genre_ids" src/test/msw/tmdb-handlers.ts` — appears only in the summary fixture and comments.
3. `pnpm test` — 485 tests green; presentation-level specs exercise the detail handler without any behaviour change.

## Tests

- [x] No new tests needed — fixtures only; the full suite is the regression check.
- [x] Coverage has not decreased (no source touched).

## Quality Gate

- [x] `pnpm format:check`
- [x] `pnpm lint` (zero warnings)
- [x] `pnpm check-types`
- [x] `pnpm test` (58 files / 485 tests)
- [x] `pnpm build`
- [x] `./scripts/check-versions.sh`

## Deployment Notes

None — test fixtures only; no env vars, no migrations, no config toggles, no runtime behaviour change.

Refs #98

### DIFF
--- a/src/test/msw/tmdb-handlers.ts
+++ b/src/test/msw/tmdb-handlers.ts
@@ -53,10 +53,13 @@ export const TMDB_MOVIE_SUMMARY_FIXTURE = {
 
 // The detail endpoint returns full `genres` objects, not the
 // numeric `genre_ids` array (that field is reserved for the
-// summary endpoints — see PR #98). Destructuring keeps the
-// detail fixture from inheriting a field `/movie/{id}` never
-// sends; each fixture mirrors its own endpoint's shape.
-const { genre_ids: _genreIds, ...summaryShape } = TMDB_MOVIE_SUMMARY_FIXTURE;
+// summary endpoints — see PR #98). Omitting it keeps the detail
+// fixture from inheriting a field `/movie/{id}` never sends;
+// each fixture mirrors its own endpoint's shape.
+const summaryShape: Omit<typeof TMDB_MOVIE_SUMMARY_FIXTURE, 'genre_ids'> = {
+  ...TMDB_MOVIE_SUMMARY_FIXTURE,
+};
+delete summaryShape.genre_ids;
 
 export const TMDB_MOVIE_DETAIL_FIXTURE = {
   ...summaryShape,
--- a/src/test/msw/tmdb-handlers.ts
+++ b/src/test/msw/tmdb-handlers.ts
@@ -51,8 +51,15 @@ export const TMDB_MOVIE_SUMMARY_FIXTURE = {
   genre_ids: [28, 12, 878],
 };
 
+// The detail endpoint returns full `genres` objects, not the
+// numeric `genre_ids` array (that field is reserved for the
+// summary endpoints — see PR #98). Destructuring keeps the
+// detail fixture from inheriting a field `/movie/{id}` never
+// sends; each fixture mirrors its own endpoint's shape.
+const { genre_ids: _genreIds, ...summaryShape } = TMDB_MOVIE_SUMMARY_FIXTURE;
+
 export const TMDB_MOVIE_DETAIL_FIXTURE = {
-  ...TMDB_MOVIE_SUMMARY_FIXTURE,
+  ...summaryShape,
   tagline: 'Your mind is the scene of the crime.',
   runtime: 148,
   genres: [
--- a/src/test/msw/tmdb-handlers.ts
+++ b/src/test/msw/tmdb-handlers.ts
@@ -53,13 +53,15 @@ export const TMDB_MOVIE_SUMMARY_FIXTURE = {
 
 // The detail endpoint returns full `genres` objects, not the
 // numeric `genre_ids` array (that field is reserved for the
-// summary endpoints — see PR #98). Omitting it keeps the detail
-// fixture from inheriting a field `/movie/{id}` never sends;
-// each fixture mirrors its own endpoint's shape.
-const summaryShape: Omit<typeof TMDB_MOVIE_SUMMARY_FIXTURE, 'genre_ids'> = {
-  ...TMDB_MOVIE_SUMMARY_FIXTURE,
-};
-delete summaryShape.genre_ids;
+// summary endpoints — see PR #98). Destructuring-rest drops
+// `genre_ids` so the detail fixture does not inherit a field
+// `/movie/{id}` never sends; each fixture mirrors its own
+// endpoint's shape.
+const {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- intentionally discarded: /movie/{id} never returns genre_ids
+  genre_ids: _discardedGenreIds,
+  ...summaryShape
+} = TMDB_MOVIE_SUMMARY_FIXTURE;
 
 export const TMDB_MOVIE_DETAIL_FIXTURE = {
   ...summaryShape,


src/test/msw/tmdb-handlers.ts
```diff
@@ -51,8 +51,18 @@ export const TMDB_MOVIE_SUMMARY_FIXTURE = {
   genre_ids: [28, 12, 878],
 };
 
-export const TMDB_MOVIE_DETAIL_FIXTURE = {
+// The detail endpoint returns full `genres` objects, not the
+// numeric `genre_ids` array (that field is reserved for the
+// summary endpoints — see PR #98). Omitting it keeps the detail
```
SebastianT2006: Functionally correct and `tsc --noEmit` is happy (the `Omit` annotation contextualises the literal and the spread widens), but the pattern is awkward: the runtime value carries `genre_ids` for one statement, the type annotation says it doesn't, and then we mutate it away. Idiomatic TS for "subset of a known shape" is destructuring rest:

```ts
const { genre_ids: _discard, ...summaryShape } = TMDB_MOVIE_SUMMARY_FIXTURE;
```

(`_discard` keeps `@typescript-eslint/no-unused-vars` quiet under the standard `argsIgnorePattern: '^_'`; if your config doesn't allow the underscore prefix, use the `void _genreIds` trick or an eslint-disable comment — your second commit title suggests you already chased this once.)

The destructuring is one line, has no type lie, no `delete`, and avoids the second commit's dance entirely. Worth folding in a follow-up if you ever touch this file again; not worth a force-push on its own for this PR.

src/test/msw/tmdb-handlers.ts
```diff
@@ -51,8 +51,18 @@ export const TMDB_MOVIE_SUMMARY_FIXTURE = {
   genre_ids: [28, 12, 878],
```
SebastianT2006, line 51: The header comment correctly explains the asymmetry and credits #98, which is the right context. Once this PR merges the `genre_ids`-only-in-summary rule is easy to remember, and the future-`.strict()` defence in the description is the real reason to land it.

SebastianT2006 (dismissed):
Clean, focused follow-up. The asymmetry is real (TMDB `/movie/{id}` returns `genres: [{id, name}]`, not `genre_ids: number[]`), and the Zod-strips-unknowns caveat in the description is accurate. Verdict: **approve, with two things to handle before merge** + a few non-blocking notes.

**Must handle before merge**

1. **`mergeable_state: blocked`** — `develop` advanced one commit beyond the PR's base (`3a…8c08` → `2aad740`): PR #100 landed between #101 being opened and now. `git fetch origin develop && git rebase origin/develop` (or merge) is enough; the CI gate that already ran was against the old base, so rebase will trigger a fresh run.
2. **Quality Gate row missing** — the table has `pnpm format:check`, `pnpm lint`, `pnpm check-types`, `pnpm test`, `./scripts/check-versions.sh`, but no `pnpm build`. CONTRIBUTING says the gate runs format → lint → types → tests → build → dependency check. Please add the build row (locally green here, but the template row should be there).

**Non-blocking, please consider**

3. **`Refs #98` in the PR footer** — the body mentions #98 in the prose but never as a reference keyword. Adding `Refs #98` (or even just `Closes #98` if you treat the follow-up as closing that PR's review thread) makes the linkage clickable in the GitHub timeline.
4. **Deployment Notes section** — empty in the body. Template permits "None", worth filling in for the audit trail.
5. **`Type of Change`** — `Chore / Maintenance` is the right pick, but `Test only` is also a checkbox option and arguably more accurate (the source tree is untouched). Either is defensible; flagging only because the repo tends to mix the two.

**Out-of-scope, tracked separately**

Filed #X — `tmdb-handlers.ts` still claims its fixtures "mirror the shape TMDB actually returns" in the file header, but `TMDB_MOVIE_SUMMARY_FIXTURE` is missing `adult`, `original_language`, `original_title`, `popularity`, `video`, and `TMDB_MOVIE_DETAIL_FIXTURE` is missing a dozen more (`belongs_to_collection`, `homepage`, `imdb_id`, `production_companies`, `spoken_languages`, `status`, etc.). Today Zod strips unknown keys so it's harmless, but it means the file header is now factually incorrect post-#101. Issue is to either narrow the header claim or flesh out the baselines — separate from this PR.

Re-ran locally on the PR head: 469/469 tests, `pnpm lint`, `pnpm check-types`, `pnpm format:check` all clean. CI `Quality gate` run on the current head: ✅ `success`. Approving once #100 is folded in.

src/test/msw/tmdb-handlers.ts
```diff
@@ -51,8 +51,20 @@ export const TMDB_MOVIE_SUMMARY_FIXTURE = {
...
 };
 
+// The detail endpoint returns full `genres` objects, not the
+// numeric `genre_ids` array (that field is reserved for the
+// summary endpoints — see PR #98). Destructuring-rest drops
+// `genre_ids` so the detail fixture does not inherit a field
+// `/movie/{id}` never sends; each fixture mirrors its own
```
SebastianT2006, line 58: Nice — exactly the pattern I suggested in the previous review. One line, no type lie, no `delete` mutation. The inline `eslint-disable-next-line` with the rationale comment is acceptable; the alternative is enabling `varsIgnorePattern: '^_'` project-wide so any `_`-prefixed binding is ignored, but a one-off disable is fine for a single binding in a test fixture.

SebastianT2006 (approved):
All five points from the previous review addressed:

1. ✅ **Branch rebase** — base is now `39ed861` (current `develop`), 0 commits behind. Up to date.
2. ✅ **`pnpm build` row** added to the Quality Gate table.
3. ✅ **`Refs #98`** keyword at the PR footer.
4. ✅ **Deployment Notes** filled in ("None — test fixtures only").
5. ✅ **Bonus**: adopted the destructuring-rest pattern I suggested (with an inline `eslint-disable-next-line` + rationale). One line, no type lie, no `delete` mutation — see the inline comment.

Re-verified locally on head `5aab7b1`:

- 485/485 tests under Node 25 simulation (consistent with the suite size after #99/#100 landed).
- 485/485 tests in plain Node 22.
- `pnpm lint` ✓, `pnpm check-types` ✓, `pnpm format:check` ✓, `pnpm build` ✓ (the rolldown chunk-size warnings are the usual noise from prior builds, not introduced here).
- Latest CI `Quality gate` run (ID 97813850815): `completed - success`.

`mergeable_state` is still reporting `blocked` in the API response, but `git rev-list origin/develop...HEAD` shows `0 3` — develop is not ahead, branch is ready. This is a stale mergeability cache that should clear on the next UI load or push; nothing actionable from the author. The deploy workflow and the GH Pages branch rule shouldn't apply here since this is a `chore/* → develop` PR.

Approving — squash and merge per `CONTRIBUTING.md` § Merge strategy. Out-of-scope follow-up tracked in #102 (header comment claims fixtures "mirror the shape TMDB actually returns" but baselines still miss a number of TMDB fields; not a blocker for this PR).

